PR #247, opened 2021-02-22
[BBS-233] Remove Sent2Vec

FrancescoCasalegno:
## Description

Same goal as #244, i.e. making code `python 3.9`-compatible, but achieve that by getting rid of `sent2vec` and `bsv`.

Fixes BBS-233. 
Closes #244.

## Todo
- [x] remove `sent2vec` and `bsv` from `src/`
- [x] remove `sent2vec` and `bsv` from `setup.py` etc.
- [x] remove `sent2vec` and `bsv` from `tests/` and `benchmarks/`
- [x] remove `sent2vec` and `bsv` from `notebooks/`

## Checklist:

- [x] This PR refers to an issue present in the [issue tracker](https://github.com/BlueBrain/Search/issues
) (if it is not the case, please create an issue
 first).
- [x] Unit tests added.
- [x] Documentation and `whatsnew.rst` updated (if needed).
- [x] `setup.py` and `requirements.txt` updated with new dependencies.
- [x] Travis CI pipeline run (add `[test travis]` to your git commit to run it)
- [x] Restore Travis CI `travis.yaml` file

PR #258, opened 2021-02-23
[BBS-291] Migrate to GitHub Actions

FrancescoCasalegno:
## Description

Removes Jenkins and Travis CI support, adds GitHub Actions support.
Fixes BBS-291.

## TODO
- [x] remove old plans
- [x] have all tests passing on GitHub actions
- [x] add secret for PyPI
- [x] merge `master` in this branch after #247  gets merged
- [x] run `lint`, `docs`, `check-packaging` only  if Python 3.7
- [x] update PR template info on triggering CI
- [x] update `CONTRIBUTING.md` info on triggering CI
- [x] cancel actions if new commit is pushed on branch

## Checklist:

- [x] This PR refers to an issue present in the [issue tracker](https://github.com/BlueBrain/Search/issues
) (if it is not the case, please create an issue
 first).
- [x] Documentation and `whatsnew.rst` updated (if needed).
- [x] Type annotations added (if a function is added or modified). 
- [x] Travis CI pipeline run (add `[test travis]` to your git commit to run it)


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+
+      - name: linting and code style
+        if: matrix.python-version == '3.7'
+        run: tox -vv -e lint
+
+      - name: tests and coverage
+        run: tox -vv -e ${{ matrix.tox-env }}
```
jankrepl, line 57:
```suggestion
        run: tox -vv -e ${{ matrix.tox-env }} -- --color=yes
```
This should give us nice colors, no? Maybe I am wrong though.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: install python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+
+      - name: linting and code style
+        if: matrix.python-version == '3.7'
+        run: tox -vv -e lint
+
+      - name: tests and coverage
```
jankrepl, line 56: Looking at the results, it seems like the `docker` / `mysql` tests are working out of the box, is that right? Cool!

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: checkout latest commit
+        uses: actions/checkout@v2
+
+      - name: get history and tags for SCM versioning to work
+        run: |
+          git fetch --prune --unshallow
+          git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+          git describe --tags
+          git describe --tags $(git rev-list --tags --max-count=1)
```
jankrepl, line 30: What exactly do we do it for?

jankrepl (commented):
Great job!!!

.github/workflows/publish-pypi.yml
FrancescoCasalegno: I had an issue with the version of the generated packages  (both `sdist` and `bdist_wheel`) that seemed related to `setuptools_scm` being unable to check which version was the right one to use. I also had the following warning
```python
UserWarning: "/home/runner/work/Search/Search" is shallow and may cause errors
0.1.dev1+g767e300
  warnings.warn('"{}" is shallow and may cause errors'.format(wd.path))
```
which led me to this [issue on GitHub](https://github.com/pypa/setuptools_scm/issues/414) where the lines of code you are highlighting now were indicated as the solution. I then tested by triggering the PyPI workflow and it was finally working fine!

.github/workflows/run-tests.yml
FrancescoCasalegno: Good idea! Done in 6f5cf1780e4f3a4313a77a07729a6a8c3bdce16f

.github/workflows/run-tests.yml
FrancescoCasalegno: It seems so!

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
+name: ci testing
+
+
+on:
+  push:
+    branches: master
+  pull_request:
+
```
pafonta, line 8: Is it expected that this section is empty?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+
+jobs:
+
+  run_test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
```
pafonta, line 17: Is there a rationale on not setting it to `true`?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+
+      - name: linting and code style
+        if: matrix.python-version == '3.7'
+        run: tox -vv -e lint -- --color=yes
+
+      - name: tests and coverage
+        run: tox -vv -e ${{ matrix.tox-env }} -- --color=yes
```
pafonta, line 57: Is `matrix.tox-env` able to find the `tox-env` values nested inside `matrix.include.[*].tox-env`?

setup.py
```diff
@@ -17,7 +17,36 @@
...
+    "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
+    "Operating System :: Unix",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3 :: Only",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Topic :: Scientific/Engineering :: Image Processing",
```
pafonta, line 42: I don't think this topic is relevant for BBS, isn't it?

pafonta (commented):
Looks cool!

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: install python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install tox
+
+      - name: linting and code style
+        if: matrix.python-version == '3.7'
+        run: tox -vv -e lint -- --color=yes
```
jankrepl, line 54: `--color=yes` is actually a `pytest` command and we just use `tox` ability to send through the CLI options. So it does not really make sense to use it for nontest `tox` environments. I think it will be just ignored.

.github/workflows/run-tests.yml
FrancescoCasalegno: Default is indeed `true`, but this means that as soon as one of the cases of the matrix that are run in parallel (i.e. `python3.7`, `python3.8`, `python3.9`) fails, all the other ones are also cancelled and marked as "failed".

I think it makes more sense to see exactly _which_ tests are failing, it may happen (we saw that recently) that e.g. we fail `3.9` but pass `3.7` and `3.8` which is more insightful than just seeing everything as "failed".

.github/workflows/run-tests.yml
FrancescoCasalegno: Yes, it means "on any pull request", afaiu.

.github/workflows/run-tests.yml
FrancescoCasalegno: Yes! I also tried with `${{ matrix.tox-env}}` but it failed—here it should be w/o brackets.

setup.py
FrancescoCasalegno: Nice catch, thank you very much! Fixed in ff77a07

.github/workflows/run-tests.yml
FrancescoCasalegno: You are right, I have to fix that!

IIUC, this is implemented by this line in `tox.ini`:
```bash
    pytest {posargs:tests}
```

Do we have reasons to prefer this over a specific section `[pytest]` in `tox.ini`?

jankrepl: IMO having it inside of `.github/workflows/run-tests.yml` is better because for example on my laptop the colors work out of the box and it is just github actions that need this extra option.

FrancescoCasalegno: Fair enough, fixed in e52753a!

.github/workflows/run-tests.yml
pafonta: > I also tried with ${{ matrix.tox-env }} but it failed

Is there a typo? Because this is what is in the file.

FrancescoCasalegno:
@jankrepl @pafonta I tried to address all your points, could you please check it's fine with you?
Thank you in advance!

pafonta:
Hello @FrancescoCasalegno !

Thanks! Had you had a look at https://github.com/BlueBrain/Search/pull/258#discussion_r582896720?

jankrepl (approved):
Great! I think we are ready to merge!

PR #263, opened 2021-03-01
[BBS-256] Harmonize logging levels

EmilieDel:
Fixes [BBS-256](https://bbpteam.epfl.ch/project/issues/browse/BBS-256).

This PR tries to harmonize the logging levels for every entrypoints.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -76,8 +76,8 @@ BBS_MINING_MYSQL_PASSWORD=guest
 #------------------------------------------------------------------------------
 # Container - mining cache creation
 #------------------------------------------------------------------------------
+BBS_MINING_CACHE_LOG_LEVEL=20
 BBS_MINING_CACHE_TARGET_TABLE=mining_cache_temporary
 BBS_MINING_CACHE_PROCESSORS_PER_MODEL=1
 BBS_MINING_CACHE_LOG_FILE=mining_cache_creation.log
-BBS_MINING_VERBOSE=-vv
 BBS_SSH_USERNAME=my_username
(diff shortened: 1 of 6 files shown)


jankrepl (approved):
Looks good:) 
BTW it seems like we are using mostly `INFO` level in our code.

```
$ git grep logger.info | wc -l
     200
$ git grep logger.debug | wc -l
      37
$ git grep logger.warning | wc -l
       7
$ git grep logger.error | wc -l 
       9
```

.env.example
```diff
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-# Integer values of Python loggin levels are as follows (default is WARNING):
+# Integer values of Python logging levels are as follows (default is WARNING):
```
pafonta, line 18:
```suggestion
# Integer values of Python logging levels are as follows (default is INFO):
```

EmilieDel: Good finding! Thanks @pafonta !

pafonta (changes requested):
Oh, sorry, I forgot:

We need to update the `README` too.

Especially the command using `--verbose` and the first part of the NB mentioning it here:
https://github.com/BlueBrain/Search/blob/master/README.md#create-the-mining-cache

EmilieDel:
Really good point @pafonta! I am changing that this afternoon! Thanks!

Stannislav:
I just noticed we have `BBS_MINING_VERBOSE=-vv` in `.env.example`. This might be related to the changes introduced in this PR.

EmilieDel:
Thanks @Stannislav, it is really relevant for this PR. I tried to address your comment in 3bd5d1e.

pafonta (approved):

PR #264, opened 2021-03-01
[BBS-305] Upgrade docker base image to python 3.7

Stannislav:
Fixes [BBS-305](https://bbpteam.epfl.ch/project/issues/browse/BBS-305)

## Description

The `bbs_base` docker image had to be re-built. In the course of this update I also made the following changes:
- Add `jupyterlab==3.0.9` to our `requirements.txt` so that it's installed on `bbs_base`
- Add `.dvc/cache` to `.dockerignore` otherwise `docker build` takes forever
- Remove a lot of configuration from `base.Dockerfile` that is now absolete
- Adjust the configuration for the new Jupyter Lab 3
- Streamline the user configuration

## How to test?

- Run `docker run --rm -Pitu $USER -v /raid:/raid bbs_base` and make sure it works
- Make sure your user name is recognised correctly
- Run `python -V` and make sure it's py37
- Install `bluesearch`
- Launch `jupyter lab` and make sure it works
- [optional] Run `pytest`
- [optional] Build the server images on top of `bbs_base` and verify that they work

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/docker/utils.sh
+++ b/docker/utils.sh
@@ -36,26 +36,27 @@ dvc_pull_models() {
 
 
 add_aliases() {
-  # Write some useful aliases to "$HOME_DIR/.bash_aliases$
-  local HOME_DIR="$1"
-
-  echo "
-  alias ll='ls -lah'\n
-  " >> "${HOME_DIR}/.bash_aliases"
+  # Write some useful aliases to "~/.bash_aliases"
+  echo "alias ll='ls -lah'\n" >> "$HOME/.bash_aliases"
 }
 
 improve_prompt() {
-  # Change the prompt appearance and add current git branch
-  local HOME_DIR="$1"
-  local USER_MODE="$2"
-  local USER_COLOR="$3"
-  if [ -z "$USER_MODE" ]; then USER_MODE="01"; fi
-  if [ -z "$USER_COLOR" ]; then USER_COLOR="33"; fi
+  # Different prompt colours for root and non-root users
+  if [ -z "$USER" ]
+  then
+    local USER_MODE="03"
+    local USER_COLOR="36"
+  else
+    local USER_MODE="01"
+    local USER_COLOR="33"
+  fi
 
+  # Define parts of PS1
   local USER_STR="\[\e[${USER_MODE};${USER_COLOR}m\]\u\[\e[00m\]"
   local WORKDIR_STR="\[\e[01;34m\]\w\[\e[00m\]"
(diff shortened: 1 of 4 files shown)


docker/utils.sh
```diff
@@ -64,35 +65,33 @@ function parse_git_branch {
...
-  local user_name="$1"
-  local home_dir="$2"
-
-  if [ "$user_name" = "root" ]
-  then
-    jupyter-lab --generate-config
-  else
-    su "$user_name" -c 'jupyter-lab --generate-config'
-  fi 
-  sed -i"" \
```
EmilieDel, line 82: Just for my understanding, why do you remove the option for the user `root` ? Same for `download_nltk` function ?

EmilieDel (approved):
LGTM! I checked by testing on DGX and seems to work fine (it is now python 3.7 and all tests are passing). Thanks for this PR @Stannislav

pafonta (commented):
Regarding `base.Dockerfile`, why rewrite what I did in https://github.com/BlueBrain/Search/pull/255? Especially because the rewrite doesn't actually add features, isn't it?

docker/utils.sh
Stannislav: Yes, that's because it was changed how the functions are called. Before they were called by root, so you had to use `su` inside the function, but now they are called directly by the appropriate user, so you don't need `su`.

Just to give a toy example:

Before:
```shell
function do_something() {
  local user_name="$1"
  if [ "$user_name" = "root" ]
  then
    echo "I'm root"
  else
    su "$user_name" -c "echo I'm $user_name"
  fi 
}

# Run do_something as root, then change to $user_name inside of do_something
user_name="<whatever>"
do_something $user_name
```

After
```shell
function do_something() {
  echo "I'm $USER"
}

# Run do_something as user $user_name straight away
user_name="<whatever>"
su $user_name -c "$(declare -f do_something) && do_something"
```

EmilieDel: I see! Thanks for the explanation @Stannislav

jankrepl (approved):
LGTM! Thank you!

pafonta (approved):
Image built from scratch:

- [x] Run docker run --rm -Pitu $USER -v /raid:/raid bbs_base and make sure it works
- [x] Make sure your user name is recognised correctly
- [x] Run python -V and make sure it's py37
- [x] Install bluesearch
- [x] Launch jupyter lab and make sure it works

PR #266, opened 2021-03-05
[BBS-282] [BBS-241] Update docs + Upload on RTD

FrancescoCasalegno:
Fixes BBS-282, BBS-241.

## Description

This PR makes our code ready for uploading on RTD.
1. Add info on installing with PyPI (i.e. `pip install bluesearch`) to the documentation and `README.md`.
1. Add a `.readthedocs.yml` configuration file for RTD.
2. Drop `mysqlclient` dependency (used only for `mysqldb` fast connections in our src code) if building on RTD. The rationale for this is as follows.
    1. `mysqlclient` needs the C library `libmysqlclient-dev` to be installed (see [here](https://stackoverflow.com/questions/47870628/oserror-mysql-config-not-found-when-trying-to-pip-install-mysqlclient-dja/51511294#51511294))
    2. RTD does not allow to install C libraries (see [here](https://docs.readthedocs.io/en/stable/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules)) and even though `libmysqlclient-dev` should be installed in the Docker image of RTD (see [here](https://github.com/readthedocs/readthedocs-docker-images/blob/master/Dockerfile#L35)) there is an known bug that makes this library unavailable (see [here](https://github.com/readthedocs/readthedocs-docker-images/issues/158)).
    3. Luckily we do not `import`the `mysqldb` module anywhere in our code (it is only used for establishing efficient connections to the MySQL servers) so we can drop it from our dependencies if we just need to build the `docs`. 
3. **Important** the `stable` build of RTD is failing. This is because the git commit we tagged with `v0.1.0` still had `mysqlclient` as a dependency. For this reason, as this PR gets merged, we should then move the tag to this commit (I know, it's awful).

## How to test?

Builds are passing and you can see the docs on RTD [here](https://readthedocs.org/projects/blue-brain-search/).

## Checklist

  (if needed)
  (if needed)
- [x] All CI tests pass.


jankrepl (approved):
Looks great! Thanks for this PR.

EmilieDel (approved):
LGTM ! Thanks for the PR and the explanation!

Stannislav (approved):
Looking good, thanks!

PR #270, opened 2021-03-10
Use PyPI badge for latest release

FrancescoCasalegno:
Tiny change, just replacing GitHub Releases badge with PyPI one.

**Rationale**: We are now publishing our package with PyPI, so there does not seem to be reasons to use GitHub Releases any more.


jankrepl (approved):
Makes sense!

pafonta (approved):
Nice :)

PR #299, opened 2021-03-15
Remove NLTK Dependencies

### LINKED ISSUE #292: Remove nltk dependency
## Context
We used to have some embedding models that used `nltk` for pre-processing, but that is not the case any more:
```bash
$ git grep -i nltk src/    |     wc -l
       0
```
So we can get rid of `nltk` from our dependencies and also remove all the `nltk.download(...)`.

Stannislav:
Fixes #292 

## Description

Remove all NLTK dependencies since we don't use it any more.

## How to test?
I did the following grep on our code base:
```bash
find . -type f ! -path './.*' ! -path './venv/*' | xargs grep -in nltk
```
Any other places NLTK could be?

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/data_and_models/pipelines/sentence_embedding/eval.py
+++ b/data_and_models/pipelines/sentence_embedding/eval.py
@@ -57,9 +57,6 @@ args = parser.parse_args()
 
 
 def main():
-    import nltk
-    nltk.download("punkt")
-
     print("Reading params.yaml...")
     params = yaml.safe_load(open("params.yaml"))["eval"][args.model]
     module = importlib.import_module("bluesearch.embedding_models")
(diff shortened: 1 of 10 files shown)


jankrepl (approved):
LGTM! Thanks!

EmilieDel (approved):
LGTM! Thanks 👍

FrancescoCasalegno (approved):
LGTM!

This PR modifies one of the dependencies of the `sentence_embedding` pipeline of `dvc`, but I think we can just decide to re-trigger `dvc` when we really want a new release and there is not a real urgency to do it now.

Unless I am wrong, 🟢  light from my side to merge into `master`!

PR #301, opened 2021-03-16
Simplify SQLAlchemy Requirements

Stannislav:
Instead of requiring

- `SQLAlchemy`
- `mysqlclient`
- `pymysql`

in `setup.py` require

- `SQLAlchemy[mysql,pymysql]`

which does exactly the same thing.


EmilieDel (approved):

jankrepl (approved):
Clever!

Issue #286, opened 2021-03-12
SearchServer and compute_embeddings should support same models


FrancescoCasalegno:
## 🐛 Bug description

The entry point `compute_embeddings` handles more models than the search server. This means that some embeddings could be computed but aren't usable. Indeed, the search server cannot load the same models to embed the queries. The models usable for search are hardcoded in `SearchServer._get_model(...)`.

The goal is to 
- [ ]remove the hardcoded model names in _get_model(...)
- [ ]  merge _get_model(...) with get_embedding_model(...) to have a unique point of truth on how to instantiate an embedding model.

## Motivation

- We cannot use all the models we can compute embeddings for.
- We need to change the code each time we want to experiment with a new model.
- Maintenance is a nightmare because the way to get a model is at several places.
- The hardcoded model names make us look unprofessional.

PR #303, opened 2021-03-16
Remove SBioBERT dedicated class

### LINKED ISSUE #275: Remove dedicated class for SBioBERT
## Context

Currently, we have a dedicated class for `SBioBERT`:
https://github.com/BlueBrain/Search/blob/c8a884f634c9e667aea2bbba087a048ccb391dd5/src/bluesearch/embedding_models.py#L117-L128

But this is not necessary since we have the class `SentTransformer`, and we could just use SBioBERT as `sbiobert = SentTransformer('gsarti/biobert-nli')` similarly to what we already do for other models:
https://github.com/BlueBrain/Search/blob/c8a884f634c9e667aea2bbba087a048ccb391dd5/src/bluesearch/embedding_models.py#L508-L514

## To do
- [ ] Remove `SBioBERT` class and use `SentTransformer()` instead.
- [ ] Make sure all tests and DVC pipelines work fine.
- [ ] Release new version for `BlueBrain/Search`.

FrancescoCasalegno:
Fixes #275 .

## Description

Previously we were using a dedicated class for SBioBERT, where in particular we defined a method for the mean pooling with attention mask prescribed [here](https://huggingface.co/sentence-transformers/bert-base-nli-mean-tokens).

This PR reduces the complexity of that solution by leveraging [`UKPLab/sentence-transformers`](https://github.com/UKPLab/sentence-transformers/) and the pretrained model [`gsarti/biobert-nli`](https://huggingface.co/gsarti/biobert-nli) through the interface provided by our `bluesearch.embedding_models.SentTransformers`.

## How to test?

The embeddings produced using `SBioBERT` did not change. You can try to run
```python
from bluesearch.embedding_models import get_embedding_model
sbiobert = get_embedding_model("SBioBERT")
sentences = [  # put whatever you want here
     "The neuron is the basic working unit of the brain.", 
     "A system of cells interlinked within one stem."
]  
preprocessed = sbiobert.preprocess_many(sentences)
embedded = sbiobert.embed_many(preprocessed)
```
using the code on `master` and on `remove_sbiobert_class` (e.g. using `pip install -e .` + `git checkout <branch_name>`).

You will see that the arrays `embedded` are the same. **Note**: preprocessed will not be the same because `sentence-transformers` don't need any extra manual pre-processing and therefore its `preprocess()` method is the identity.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/embedding_models.py
+++ b/src/bluesearch/embedding_models.py
@@ -298,7 +298,7 @@ def compute_database_embeddings(connection, model, indices, batch_size=10):
 
 def get_embedding_model(
     model_name_or_class: str,
-    checkpoint_path: Optional[Union[pathlib.Path, str, None]] = None,
+    checkpoint_path: Optional[Union[pathlib.Path, str]] = None,
     device: Optional[str] = "cpu",
 ) -> EmbeddingModel:
     """Load a sentence embedding model from its name or its class and checkpoint.
--- a/src/bluesearch/embedding_models.py
+++ b/src/bluesearch/embedding_models.py
@@ -21,13 +21,12 @@ import logging
 import multiprocessing as mp
 import pathlib
 from abc import ABC, abstractmethod
+from typing import Optional, Union
 
 import joblib
 import numpy as np
 import sentence_transformers
 import sqlalchemy
-import torch
-from transformers import AutoModel, AutoTokenizer
 
 from .sql import retrieve_sentences_from_sentence_ids
 from .utils import H5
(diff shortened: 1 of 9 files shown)


tests/test_server/test_search_server.py
```diff
@@ -33,15 +33,14 @@ def search_client(
     """Fixture to create a client for mining_server."""
 
     sbiobert_model_inst = Mock()
-    sbiobert_model_class = Mock()
-    sbiobert_model_class.return_value = sbiobert_model_inst
     sbiobert_model_inst.preprocess.return_value = "hello"
```
jankrepl, line 36: Maybe we can change the variable name to something different? Might be confusing to still use `sbiobert_model..`

src/bluesearch/embedding_models.py
```diff
@@ -492,16 +321,17 @@ def get_embedding_model(model_name_or_class, checkpoint_path=None, device=None):
 
     Parameters
     ----------
-    model_name_or_class : str
+    model_name_or_class :
```
jankrepl, line 324: So I guess you removed the type information from the docstring on purpose since we have annotations, right?
In that case should we maybe remove the `:`? I am not sure what the convention is.

jankrepl (approved):
Nice job! Always refreshing to see deleted lines > added lines:)

tests/test_server/test_search_server.py
FrancescoCasalegno: Renamed the variable as `fake_embedding_model` in 7a66438.

src/bluesearch/embedding_models.py
FrancescoCasalegno: Fixed in 1f6eec6. By the way, I noticed that it is necessary to specify the return type of the function in the docstring or Sphinx will get confused. 

For instance, if we set the docstring
```python
   """
    Returns
    -------
    sentence_embedding_model : EmbeddingModel
        The sentence embedding model instance.
   """
```
we correctly get 
Screenshot 2021-03-17 at 14 51 15

but if we set the docstring as 
```python
   """
    Returns
    -------
    sentence_embedding_model
        The sentence embedding model instance.
   """
```
we get instead
Screenshot 2021-03-17 at 14 51 53


@jankrepl so any idea on how to treat the return type in docstrings?

jankrepl: My guess is that there are two valid options
1. `variable name : type`
2. `type`

That is why Sphinx did what it did. Maybe there is another syntax that allows for variable names only.

src/bluesearch/embedding_models.py
```diff
@@ -471,7 +296,11 @@ def compute_database_embeddings(connection, model, indices, batch_size=10):
     return final_embeddings, retrieved_indices
 
 
-def get_embedding_model(model_name_or_class, checkpoint_path=None, device=None):
+def get_embedding_model(
+    model_name_or_class: str,
+    checkpoint_path: Optional[Union[pathlib.Path, str, None]] = None,
```
pafonta, line 301:
```suggestion
    checkpoint_path: Optional[Union[pathlib.Path, str]] = None,
```

src/bluesearch/entrypoint/embedding_server.py
```diff
@@ -43,10 +43,9 @@ def get_embedding_app():
 
     # Load embedding models
     logger.info("Loading embedding models")
+    supported_models = ["SBERT", "SBioBERT", "BioBERT NLI+STS"]
     embedding_models = {
-        "SBERT": SentTransformer("bert-base-nli-mean-tokens"),
-        "BIOBERT NLI+STS": SentTransformer("clagator/biobert_v1.1_pubmed_nli_sts"),
-        "SBioBERT": SBioBERT(),
+        model_name: get_embedding_model(model_name) for model_name in supported_models
```
pafonta, line 48: Nice! I like that!

src/bluesearch/server/search_server.py
```diff
@@ -110,32 +110,30 @@ def __init__(
...
         ----------
-        model_name : str
+        model_name :
             The name of the model.
 
         Returns
         -------
-        bluesearch.embedding_models.EmbeddingModel
-            The embedding model class.
+        sentence_embedding_model :
```
pafonta, line 123: Should it be `sentence_embedding_model : EmbeddingModel` like in `embedding_models.py`?

src/bluesearch/server/search_server.py
```diff
@@ -110,32 +110,30 @@ def __init__(
...
-            ),
-            "BioBERT NLI+STS CORD-19 v1": lambda: SentTransformer(
-                biobert_nli_sts_cord19_v1_model_path
+            "SBioBERT": lambda: get_embedding_model("SBioBERT"),
+            "SBERT": lambda: get_embedding_model("SBERT"),
+            "BioBERT NLI+STS": lambda: get_embedding_model("BioBERT NLI+STS"),
+            "BioBERT NLI+STS CORD-19 v1": lambda: get_embedding_model(
+                "SentTransformer", checkpoint_path=biobert_nli_sts_cord19_v1_model_path
             ),
         }
```
pafonta, line 138: Note as a reminder:

I feel like we are almost to https://github.com/BlueBrain/Search/issues/286 but there is still an extra step to do.

So, https://github.com/BlueBrain/Search/issues/286 is still to do (not the goal of the current PR).

pafonta (approved):
Thank you for doing this!

I like that we reach more harmonized code.

There are two small changes (typing + docstring inconsistency) to do IMO before merging but otherwise I have tried the `how to test?` and it works:

```
>>> np.allclose(embedded_master, embedded_pullrequest, rtol=0, atol=1e-06)
True
```

src/bluesearch/embedding_models.py
jankrepl: Hmm, I am not sure if I get this change, could you explain?

jankrepl: Ahh, nice find!

src/bluesearch/server/search_server.py
FrancescoCasalegno: Yes I was thinking about the same.

src/bluesearch/server/search_server.py
FrancescoCasalegno: Yes, nice catch!

FrancescoCasalegno: Done in 5203164

Issue #310, opened 2021-03-17
Migrate add_er to spacy train and config.cfg


pafonta:
## 🚀 Feature

Migrate the addition of the rule-based NER components (`EntityRuler`) to `spacy train` and `config.cfg`.

## Motivation

We started in #274 to move to standardized NER workflows. spaCy includes a standard way of adding `EntityRuler` components to NER models: `[components]` in `config.cfg`.

Also, doing so will let take into account the rules during the training of the model. This might improve the performances of the resulting NER model. Note: It might have an impact on #294 and #295.

## Pitch

- Load directly the NER model(s) instead of loading them and then adding the entity rulers.
- Remove the mentions of `add_er_*` in READMEs.
- Remove the DVC stages for `add_er_*`.
- Remove `add_er.py`.

## Additional context

Reference: https://spacy.io/usage/training#config-components

Issue #308, opened 2021-03-17
Combine "spacy evaluate" with "bluesearch.mining.eval"


FrancescoCasalegno:
Currently our evaluation step in the `dvc` pipeline dedicated to NER models relies entirely on the following script, which calls functions from `bluesearch.mining.eval`.
https://github.com/BlueBrain/Search/blob/fa1331c98c8823ec85c5b3d92d58e99ab6010574/data_and_models/pipelines/ner/eval.py#L1

It could be convenient to use instead the interface provided by spaCy's CLI, i.e. [`spacy evaluate`](https://spacy.io/api/cli#evaluate). In particular, we would like that to potentially call our own evaluation functions (e.g. token-based and entity-based F1, Prec, Rec, ...). 

- [ ] does `spacy evaluate` produce the same results (entity-based F1, Prec, Rec) than our `bluesearch.mining.eval` functions? if not, why? 
- [ ] can we configure `spacy evaluate` in such a way that the `spacy.Scorer` calls our own user defined scoring functions?
- [ ] if we then don't need the annotations in JSONL format anymore, then (i) get rid of the `preprocessing` step and script (ii) use `.spacy` binaries instead of `.jsonl`.

pafonta:
UPDATE Actions

The section for evaluation in `params.yaml` could be simplified after #351.

There is indeed no more the need for mapping` scispaCy` labels to our labels.

PR #324, opened 2021-03-29
#315 Remove the boilerplate in dvc.yaml thanks to DVC 2 foreach

### LINKED ISSUE #315: Remove the boilerplate in dvc.yaml thanks to DVC 2 foreach
## 🚀 Feature

After #273 is merged:

Remove the boilerplate in `dvc.yaml` thanks to the new feature `foreach` from DVC 2 .

## Motivation

Stages of a given type (e.g. `eval_*`) use the same logic in `dvc.yaml`. The only little variation is on the arguments.

This is **error-prone** and **painful** to maintain and review DVC pipelines.

## Pitch

Thanks to DVC 2 foreach, we could regroup stages with the same logic and just have the arguments change.

## Additional context

Reference: https://dvc.org/doc/user-guide/project-structure/pipelines-files#foreach-stages.

EmilieDel:
Fixes #315.

## Description

From #315: 

> Stages of a given type (e.g. eval_*) use the same logic in dvc.yaml. The only little variation is on the arguments.
> This is error-prone and painful to maintain and review DVC pipelines.
> Thanks to DVC 2 foreach, we could regroup stages with the same logic and just have the arguments change.

During this PR:
- `NER` pipelines are simplified:
  - `training` and `eval` steps are using `foreach` feature. 
  - `add_er` is still not using `foreach`. The main reason is that the number of `etypes` is changing through models and the use of `foreach` would need a bigger refactoring. At the same time, it seems this step is going to disappear in the near future. See [comment](https://github.com/BlueBrain/Search/pull/324#issuecomment-810058603) from @pafonta. 
  - To avoid parameters duplication (e.g. `disease` and `DISEASE`), `params.yaml` was slightly changed for lowercase names instead of uppercase ones (`add_er.py` and `eval.py` were modified in consequences).  
- `sentence_embeddings` pipelines are simplified:
  - `training` and `evaluation` steps are using `foreach` feature. 
  - To make this more generic: 
    - All evaluation steps are depending on the `Dockerfile` (it was not the case for `biobert_nli_sts_cord19_v1` before) 
    - Some models had as dependency the model path which is now removed because also present in the `params.yaml` (it was not the case for `count` and `tf_idf`)

## How to test

- Changes in `dvc.yaml` seem complicated to read. The easiest is maybe to read those raw files instead. 
- Changes in `dvc.lock` are more readable. The main changes should be 
  - The stages name. 
  - Some dependencies in `sentence_embeddings` pipeline (as the `Dockerfile` and the `model` as explained above)
  - Some etype in lowercase instead of uppercase
  - md5 for `eval.py` and `add_er.py` as changes were done in consequence of the lowercase etypes

## Checklist

- [X] Unit tests added.
  (if needed)
  (if needed)
- [X] `setup.py` and `requirements.txt` updated with new dependencies.
  (if needed)
- [X] Type annotations added.
- [X] All CI tests pass.


pafonta:
Hello @EmilieDel !

Regarding not using `foreach` for `add_er_*` stages, here are the two tickets we discussed:

1 - `add_er_*` stages would be removed with the integration of the `EntityRuler` in `config.cfg`: #310.

2 - Handling each entity in `params` would be removed thanks to `spacy evaluate` and its `ents_per_type`: #308.

FrancescoCasalegno:
Conflict with PR #248 . 

Let's wait for that PR to get merged first and then we can review and merge this one.

pafonta (approved):
Thank you very much @EmilieDel for simplifying things like this!

Before merging, I would suggest to keep the comment in `ner/dvc.yaml` before `convert_annotations`. Indeed, it explains why there are theses conversions.
```
 # https://github.com/BlueBrain/Search/issues/308 would remove the need of
 # converting the annotations by dropping the .jsonl format for the .spacy one.
```

data_and_models/pipelines/ner/dvc.lock
```diff
@@ -139,20 +19,20 @@ stages:
...
     - path: eval.py
-      md5: 3c9e86f7cbfaea90929e96a6443a16d4
-      size: 3391
+      md5: 8b7c9e4c20e6b0dd7f7daa0097be8e1a
+      size: 3407
     params:
       params.yaml:
-        eval.DISEASE:
+        eval.disease:
           etype_name: DISEASE
```
FrancescoCasalegno, line 27: I see that entity types got consistently renamed as lowercase. Is there a reason for this?
Also notice that the `etype_name` just below is uppercase, which seems contrasting.

FrancescoCasalegno (commented):
Thanks for doing this, looks fine to me!

I have a question about uppercase/lowercase, but once fixed I would give my green light.

jankrepl (approved):
Wow, seems really useful in our setup. In general, not sure how I feel about turning yaml files into scripting languages but I guess we discussed that already:)

data_and_models/pipelines/ner/dvc.lock
Stannislav: Also, hijacking this thread to as a related question. Is there a reason why we write
```yaml
CHEMICAL:
  etype_name: CHEBI
```
rather than
```yaml
CHEMICAL: CHEBI
```
I know it was not introduced in this PR, I just never thought about it before.

FrancescoCasalegno:
#248 got merged, so this issue is not **blocked** any more.

data_and_models/pipelines/ner/dvc.yaml
```diff
@@ -1,404 +1,170 @@
...
+      annotation11:
+        annotation_filename: annotations11_CharlotteLorin_2020-08-28_raw1_10EntityTypes
+      annotation12:
+        annotation_filename: annotations12_EmmanuelleLogette_2020-08-28_raw7_10EntityTypes
+      annotation13:
+        annotation_filename: annotations13_CharlotteLorin_2020-09-02_raw7_10EntityTypes
+      annotation14:
+        annotation_filename: annotations14_EmmanuelleLogette_2020-09-02_raw8_CellCompartmentDrugOrgan
+      annotation15:
+        annotation_filename: annotations15_EmmanuelleLogette_2020-09-22_raw9_Pathway
```
Stannislav, line 21: I'm wondering if the names `annotationXX` are used anywhere? Are they used at all? On the [DVC documentation](https://dvc.org/doc/user-guide/project-structure/pipelines-files#foreach-stages) I read that it's possible to write something like
```yaml
stages:
  convert_annotations:
    foreach:
      - annotations5_EmmanuelleLogette_2020-06-30_raw2_Disease
      - annotations6_EmmanuelleLogette_2020-07-07_raw4_TaxonChebi
      - ...
    do:
      cmd: python preprocess.py ../../annotations/ner/${item}.jsonl
      ...
```
Would we break anything if we used this syntax?

data_and_models/pipelines/ner/dvc.yaml
```diff
@@ -1,404 +1,170 @@
...
+         base_model: en_ner_bionlp13cg_md
+       model3:
+         annotation_filename: annotations6_EmmanuelleLogette_2020-07-07_raw4_TaxonChebi
+         base_model: en_ner_craft_md
+       model4:
+         annotation_filename: annotations9_EmmanuelleLogette_2020-07-08_raw6_CelltypeProtein
+         base_model: en_ner_jnlpba_md
+       model5:
+         annotation_filename: annotations15_EmmanuelleLogette_2020-09-22_raw9_Pathway
+         base_model: en_ner_craft_md
```
Stannislav, line 50: I don't know if it's possible, probably it's a discussion for another PR/issue, but maybe someone has some thoughts/insight on this already.

I see that in all of the stages we use similar data about the models, their entity types, and the corresponding annotations. It feels to me that this might be better  stored in `params.yaml` (correct me if I'm wrong!) Apparently one can iterate over dictionaries from `params.yaml` in the `dvc.yaml` file, here's the example from the [DVC docs](https://dvc.org/doc/user-guide/project-structure/pipelines-files#foreach-stages):
```yaml
stages:
  mystages:
    foreach: ${myobject} # From params.yaml
    do:
      cmd: ./script.py ${key} ${item.prop1}
      outs:
        - ${item.prop2}
```
I'm just thinking out loud, but couldn't we put something like this in the `params.yaml` file and then use it in `dvc.yaml` like in the example above:
```yaml
#params.yaml
...
models:
  model1:
    annotation_filename: annotations5_EmmanuelleLogette_2020-06-30_raw2_Disease
    base_model: en_ner_bc5cdr_md
    entity_types:
      DISEASE: DISEASE
  model2:
    annotation_filename: annotations14_EmmanuelleLogette_2020-09-02_raw8_CellCompartmentDrugOrgan
    base_model: en_ner_bionlp13cg_md
    entity_types:
      CELL_COMPARTMENT: CELLULAR_COMPONENT
      DRUG: SIMPLE_CHEMICAL
      ORGAN: ORGAN
...
```

data_and_models/pipelines/ner/dvc.yaml
```diff
@@ -417,4 +183,4 @@ stages:
     - ../../metrics/ner/interrater/organ.json
     - ../../metrics/ner/interrater/organism.json
     - ../../metrics/ner/interrater/pathway.json
-    - ../../metrics/ner/interrater/protein.json
+    - ../../metrics/ner/interrater/protein.json
```
Stannislav, line 186: Can we restore the new line at the end of file here 😅

Stannislav (approved):
Nice, I think it looks much better now :) I left a couple of comments here and there

data_and_models/pipelines/ner/dvc.lock
EmilieDel: The main idea was to avoid duplication of `parameters` in the `dvc.yaml`.  Indeed, `params.yaml` keys are expecting uppercase names and `metrics` filenames are lowercase. I could not find a way (maybe there is one) to modify directly the strings in the `dvc.yaml` from one form to the other. From what @pafonta told me, `spaCy 3.0` allows to do this conversion of `entity types` directly in the model. So I thought it is better to change `params.yaml` that is going to change anyway in the future (see comment from [PA](https://github.com/BlueBrain/Search/pull/324#issuecomment-810058603)).

However, here are the different possibilities (in my opinion):
- Having 2 parameters in `dvc.yaml` for entity types (lower and uppercase), or finding a way to changing strings directly in the `dvc.yaml` if possible. (This solution does not affect the rest of the files)
- Changing the `params.yaml` keys to lowercase (as done here - affects `eval.py` and `params.yaml`)
- Going for metric filename that are uppercase instead of lowercase. 

What do you think about them @Stannislav @FrancescoCasalegno ? Do not hesitate if you have other ideas for this. 

Regarding the idea: 
```yaml
CHEMICAL: CHEBI
```
I maybe need to think about it to be sure there is no problem but it is definitely worth to check. Thanks for the suggestion!

data_and_models/pipelines/ner/params.yaml
```diff
@@ -3,21 +3,21 @@ train:
...
+  cell_type:
     etype_name: CELL_TYPE
-  PROTEIN:
+  protein:
     etype_name: PROTEIN
-  PATHWAY:
-    etype_name: PATHWAY
\ No newline at end of file
+  pathway:
+    etype_name: PATHWAY
```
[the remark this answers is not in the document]
EmilieDel: I tried to answer this question in [this comment](https://github.com/BlueBrain/Search/pull/324#discussion_r608419741).

data_and_models/pipelines/ner/dvc.yaml
EmilieDel: That is a really good point. This only thing is that the naming in the `dvc.lock` is going to become `convert_annotations@0`, `convert_annotations@1`, ... but it is not really a problem.

data_and_models/pipelines/ner/dvc.yaml
EmilieDel: Thanks @Stannislav for this suggestion! Maybe just some points to keep in mind for this discussion:
- Initially, `params.yaml` was created to keep track of the `hyperparameters` along the pipeline (as `n_epochs`, `dropout`, ...). There is indeed a DVC command `dvc params diff` allowing to do the comparison between several runs with different parameters. I have the feeling that the `dvc.yaml` is the file used to create the "links" between the steps/dependencies (DAG). So from your example, in my opinion, the `base model` could be a `hyperparameter`, I am not sure about the `annotation filename`. 
- I am wondering if changing one key of this dictionary `model1` could affect steps that are not changing but depends on this dictionary anyway. For example, you want to change the `base model`, the step `convert_annotation` is not depending on `base model` but its object changes. I am not sure how DVC handles this. 

In general, I also have the feeling that those structures could be re-think with the `config.cfg` introduced by `spaCy 3.0` as most of the hyper-parameters seems to be located in that file now.

data_and_models/pipelines/ner/dvc.lock
pafonta: > From what @pafonta told me, spaCy 3.0 allows to do this conversion of entity types directly in the model.

The idea would be to rename the labels directly in the pipeline.

One could add an `EntityRuler` in `config.cfg` (related to #310) with a `ENT_TYPE` pattern for each label.

This way, `nlp(text).ents` would have our labels only and free us from managing them outside the `spaCy` pipeline.

data_and_models/pipelines/ner/dvc.yaml
EmilieDel: Done in 60cdbde

data_and_models/pipelines/ner/dvc.lock
Stannislav: @EmilieDel since you're saying that `dvc.yaml` and `params.yaml` will change anyway, maybe the upper/lower case discussion is not so important.

Personally I like things spelled out in the most direct way (like I wrote above `CHEMICAL: CHEBI`), but this kind of restructuring might be out of scope for this PR.

FrancescoCasalegno (approved):

data_and_models/pipelines/ner/dvc.yaml
Stannislav: Thanks for the comments, I see now that maybe `params.yaml` might not be the best place for this, as it has a specific use of specifying hyperparameters.

Regarding changing some of the keys and re-computing the steps in the DAG. I'd assume that the structrue above is absolutely fixed and we'd never change it anyway. But yes, I also agree that for example the data conversion should not care about what model the data files are used for.

Anyway, this comment was not meant to lead to changes in this PR, just for duscussing ideas :) feel free to carry on :)

PR #326, opened 2021-03-31
Refactor dependencies

### LINKED ISSUE #325: Put dependencies of data_and_models in a separate place
## 🚀 Feature

We should separate the dependencies of the various pipelines and scripts of `data_and_models` from the `install_requires.

## Motivation

This will mae `pip install bluesearch` run faster, and also our tests should run faster, since the section `dev` of `extra_requires` won't be affected.

## Pitch

Create dedicated section inside of `extra_requires` for dependencies of `data_and_models`.

FrancescoCasalegno:
Fixes #325 .

## Description

1. Dependencies that are not strictly need to use  the `bluesearch` Python package are removed from `install_requires`.
2. Further dependencies that required to run the code of `data_and_models/` are added to `extra_requires`.

## How to test?

```bash
git grep -hrIE "(^ *import |^ *from \S+ import)" src | \
 grep -v " \." | \
 cut -d "." -f -1 | \
 cut -d " " -f 2 | \
 sort -u | \
 grep . > \
 deps_src.txt


git grep -hrIE "(^ *import |^ *from \S+ import)" data_and_models | \
 grep -v " \." | \
 cut -d "." -f -1 | \
 cut -d " " -f 2 | \
 sort -u | \
 grep . > \
 deps_dam.txt

echo 'import pandas as pd

s_src = set(pd.read_csv("deps_src.txt", names=["x"], header=None)["x"].to_list())
s_dam = set(pd.read_csv("deps_dam.txt", names=["x"], header=None)["x"].to_list())
s_std37 = set(pd.read_csv("deps_std37.txt", names=["x"], header=None)["x"].to_list())

print("=======   requirements for src/   =======")
print("\\n".join(sorted(s_src - s_std37)))
print()

print("=======   requirements for data_and_model/   =======")
print("\\n".join(sorted(s_dam - s_src - s_std37)))
' > uq_imports.py

python uq_imports.py
```
#### Dependencies of  `src/`
These are the dependencies that should be listed for `install_requires`. The ✅ are just to remind myself that they are included in the `setup.py`.
```bash
IPython - ✅ (package is called: ipython)
dotenv - ✅ (package is called: python-dotenv)
flask - ✅ (package is called: Flask)
h5py - ✅ 
ipywidgets - ✅ 
joblib - ✅ 
langdetect - ✅ 
matplotlib - ✅ 
numpy - ✅ 
pandas - ✅ 
pdfkit - ✅ 
pkg_resources - ✅  (included inside setuptools, which is required to pip-install bluesearch)
requests - ✅ 
sentence_transformers - ✅ 
sklearn - ✅ (package is called: scikit-learn)
spacy - ✅ 
sqlalchemy- ✅ (package is called: SQLAlchemy)
torch - ✅  
tqdm - ✅ 
```
We also add `scispacy` since most of the time `bluesearch` works with `scispacy` models.

Also, we need to add `cryptography` or the MySQL tests will not pass—maybe `cryptography` was included in one of the dependencies that got moved to `extra_requires.data_and_models`?

#### Dependencies of `data_and_models/`
For `extra_requires.data_and_models` this are the dependencies found.
```bash
prodigy - ✅ 
scipy - ✅ 
transformers - ✅ (sentence_transformer is needed in our src/, but not transformers)
yaml - ✅ (package is called: PyYAML)
```
We also need to add `dvc`, because even if we never `import` it we are using its cli commands.

## Checklist

  (if needed)
  (if needed)
- [x] All CI tests pass.


EmilieDel (approved):
LGTM! It is a nice cleaning! 

Regarding the libraries mentioned in the previous comment: 
- `joblib`: I don't have a strong opinion about this one. The `sklearn` models you are talking about is mainly `tf_idf`, right ? Do you think it is useful ? 
- `sklearn`: I don't mind keeping it in the `install_requires`. 
(However, that makes me think, that the module `src/bluesearch/mining/eval.py` should maybe be placed under `data_and_models` as it is more part of the `training/evalutation` step in our pipeline and `src` could be seen as the usage part of our model and I am not sure it is even called somewhere in `src`. It is probably too extreme, especially because then we will have problems with testing this part of code. But this is definitely one of the package making the separation between `data_and_models` and `src` complicated.)
- `tqdm`: it is a good idea to remove it and replace by a `print` statement to show the evolution. At the same time, it is not a huge package, so we can also keep it. 

What about those ones ? 
- `pkg_resources`: I don't understand why we need this one ?
- `matplotlib`: in the `src/` part, it is only used once for plotting the `confusion` matrix. (cf `plot_ner_confusion_matrix` in `src/mining/eval.py`)
- `pdfkit`: I know this is the package used to create the reports of the `Search widget`. I am wondering if it is still an useful feature.

FrancescoCasalegno:
> LGTM! It is a nice cleaning!
> 
> Regarding the libraries mentioned in the previous comment:
> 
> * `joblib`: I don't have a strong opinion about this one. The `sklearn` models you are talking about is mainly `tf_idf`, right ? Do you think it is useful ?
> * `sklearn`: I don't mind keeping it in the `install_requires`.
>   (However, that makes me think, that the module `src/bluesearch/mining/eval.py` should maybe be placed under `data_and_models` as it is more part of the `training/evalutation` step in our pipeline and `src` could be seen as the usage part of our model and I am not sure it is even called somewhere in `src`. It is probably too extreme, especially because then we will have problems with testing this part of code. But this is definitely one of the package making the separation between `data_and_models` and `src` complicated.)
> * `tqdm`: it is a good idea to remove it and replace by a `print` statement to show the evolution. At the same time, it is not a huge package, so we can also keep it.
> 
> What about those ones ?
> 
> * `pkg_resources`: I don't understand why we need this one ?
> * `matplotlib`: in the `src/` part, it is only used once for plotting the `confusion` matrix. (cf `plot_ner_confusion_matrix` in `src/mining/eval.py`)
> * `pdfkit`: I know this is the package used to create the reports of the `Search widget`. I am wondering if it is still an useful feature.

Hello @EmilieDel  thanks for your review! I realised that maybe we can separate this discussion on dependencies to remove into another issue, so for the follow-up I suggest to discuss it in the newly created #330.

To quickly answer about `pkg_resources`, it is a module distributed with `setuptools` (see [here](https://setuptools.readthedocs.io/en/latest/pkg_resources.html)) that we use in one place to get the filesystem path to a given `.css` file using `pkg_resources.resource_filename` (see [here](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resource-extraction)).

Stannislav (approved):
Yes, looks good to me!

I'm also in favour of dropping `joblib` and `tqdm` in the way you suggested.

Regarding `scikit-learn` and the confusion matrix. We can leave it if it's too much hassle to re-implement it. I found a suggestion of how to do it using `seaborn` [here](https://stackoverflow.com/a/42265865/2804645), it's probably possible to do something similar in pure `matplotlib`.

pafonta (approved):
Happy to see that the separation between the code and data+models takes form.

Before merging, `prodigy` is to be removed. Indeed, this is no longer a dependency since `v0.1.2`.

jankrepl (approved):
Looks great! Thank you!

FrancescoCasalegno:
> Before merging, `prodigy` is to be removed. Indeed, this is no longer a dependency since `v0.1.2`.

Done in 2b08a6c, thanks for pointing it out!

Issue #319, opened 2021-03-23
Train NER model for "PATHWAY" using spaCy


FrancescoCasalegno:
1. Use frozen backbone corresponding to the BioBERT_NLI_STS_CORD19_v1.
2. Use training set = annotations15_EmmanuelleLogette_2020-09-22_raw9_Pathway.jsonl.

**Note**: it's the same of #318 , just using `spacy` 's 2.0 `spacy train` command instead of 🤗's `transformers` interface.

pafonta:
Closed.

The experiment was done in #328.

It was closed to deploy it for all entities: https://github.com/BlueBrain/Search/pull/328#issuecomment-818929155.

PR #329, opened 2021-04-01
#318 Train NER model for "PATHWAY" using 🤗 /transformers 

### LINKED ISSUE #318: Train NER model for "PATHWAY" using 🤗 /transformers
1. Use frozen backbone corresponding to the `BioBERT_NLI_STS_CORD19_v1`.
2. Use training set = `annotations15_EmmanuelleLogette_2020-09-22_raw9_Pathway.jsonl`.

**Note**: it's the same of #319 , just using 🤗's `transformers` interface instead of  `spacy` 's 2.0 `spacy train` command.

EmilieDel:
Fixes #318.

## Description

Train NER model for "PATHWAY" using 🤗 /transformers . 

## How to test?

All scripts are under `data_and_models/pipelines/ner/transformers_vs_spacy/`
To train and evaluate `transformers`:
```
cd data_and_models/pipelines/ner/transformers_vs_spacy/transformers/
./0_prepare_data.sh                       # outputs test_data.txt & train_data.txt
./1_run_transformers_ner.sh               # outputs test_data.pkl & train_data.pkl
./2_eval_pred.sh                          # outputs pathway_metrics_token.json & pathway_metrics_entity.json
python 3_compare_tokens.py                # outputs analysis on the set of train and test entities
```
To evaluate `spacy`:
```
cd data_and_models/pipelines/ner/transformers_vs_spacy/spacy/
./eval.sh                                 # outputs pathway_metrics.json 
python compare_tokens.py                  # outputs analysis on the set of train and test entities
```

## Results: Metrics and Comparison of entitities sets

You can reproduce all those results following steps of **How to test?** section. 
- For `spaCy` (the model evaluated is the one before the step of `entity ruler`, even if the results don't change):
  - Metrics (token-level):
  ```{"precision": 0.7090909090909091, "recall": 0.4854771784232365, "f1-score": 0.5763546798029556, "support": 241}```
  - Tokens analysis:
  ```
  len(train) = 91
  len(test) = 132
  len(pred) = 78
  
  VENN: train vs. test
  [ 47 | 44 | 88 ]
  VENN: train vs. pred
  [ 47 | 44 | 34 ]
  VENN: test vs. pred
  [ 75 | 57 | 21 ]
  
  How many of the unseen tokens were predicted?
  Out of 88 unseen tokens 16 were predicted
  Out of 44 seen tokens 41 were predicted
  ```

- For `transformers`:
  - Metrics (token-level):
  ```{'precision': 0.5895196506550219, 'recall': 0.5601659751037344, 'f1-score': 0.574468085106383, 'support': 241}```

  - Token Analysis:
  ``` 
  len(train) = 91
  len(test) = 132
  len(pred) = 127
  
  VENN: train vs. test
  [ 47 | 44 | 88 ]
  VENN: train vs. pred
  [ 44 | 47 | 80 ]
  VENN: test vs. pred
  [ 63 | 69 | 58 ]
  
  How many of the unseen tokens were predicted?
  Out of 88 unseen tokens 29 were predicted
  Out of 44 seen tokens 40 were predicted
  ```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


FrancescoCasalegno:
Hi guys, I have a few questions.
1. Should it run on GPU or also locally is also fine? Should we expect very long runtimes?
2. If you run these experiments it in a Docker container, could you please add to "How to Test" also the commands used to build the Docker image + to run the Docker container?
3. Are training results supposed to be reproducible?
3. This `seqeval` seems to implement many of the features of `bluesearch.mining.eval`. Did you check if the outputs you obtain using the two are the same? Should we consider dropping `bluesearch.mining.eval`?

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/1_run_transformers_ner.py
```diff
@@ -0,0 +1,526 @@
...
+    #model = OurBertClassifier(model_args.model_name_or_path, config=config)
+    # Freeze the BERT parameters
+    for param in model.bert.parameters():
+        param.requires_grad = False
+    for n_layer in range(12, 24):
+        for param in model.bert.encoder.layer[n_layer].parameters():
+            param.requires_grad = True
+    n_trainable = sum(param.numel() for param in model.parameters() if param.requires_grad == True)
+    print("Number of trainable parameters:", n_trainable)
+    input("Press any key to continue")
```
jankrepl, line 327: I think what we forgot about here is that there are the fixed vocab embeddings that are at the very start of the architecture. Under this setup they are frozen. See
```
https://github.com/huggingface/transformers/blob/76800fb8e66aabfa1b31300f80fcf986befbb15f/src/transformers/models/bert/modeling_bert.py#L846
```

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/1_run_transformers_ner.py
```diff
@@ -0,0 +1,526 @@
...
+        use_auth_token=True if model_args.use_auth_token else None,
+    )
+    #model = OurBertClassifier(model_args.model_name_or_path, config=config)
+    # Freeze the BERT parameters
+    for param in model.bert.parameters():
+        param.requires_grad = False
+    for n_layer in range(12, 24):
+        for param in model.bert.encoder.layer[n_layer].parameters():
+            param.requires_grad = True
+    n_trainable = sum(param.numel() for param in model.parameters() if param.requires_grad == True)
```
jankrepl, line 325:
```suggestion
    n_trainable = sum(param.numel() for param in model.parameters() if param.requires_grad)
```
This should be enough:)

Stannislav:
Maybe it can also be interesting to try and fine tune our `biobert_nli_sts_cord19_v1` model from `data_and_models/models/sentence_embedding/biobert_nli_sts_cord19_v1` like in #319

pafonta (commented):
I tried to follow the instructions to get the model.

Are the errors below 'expected'?

After doing:
```
git checkout transformers_vs_spacy
dvc pull data_and_models/annotations/ner/*.dvc
cd data_and_models/pipelines/ner/transformers_vs_spacy/transformers/
./0_prepare_data.sh
```

I get lots of the following errors:

```
Something went wrong: "

Something went wrong:                             " O

Something went wrong: "

Something went wrong:                             " O
```

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/1_run_transformers_ner.sh
```diff
@@ -0,0 +1,43 @@
...
+python3 1_run_transformers_ner.py \
+  --model_name_or_path "dmis-lab/biobert-large-cased-v1.1" \
+  --output_dir "checkpoints/$exp_name" \
+  --do_train \
+  --do_eval \
+  --do_predict \
+  --evaluation_strategy "steps" \
+  --eval_steps 10 \
+  --train_file "$DS" \
+  --validation_file "$DS_VAL" \
```
pafonta, line 36: Is it expected that the `test` data are used as `dev` data in this PR?

This isn't what we are doing at the moment.

The reported performances are therefore not comparable.

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/2_eval_pred.py
```diff
@@ -0,0 +1,97 @@
...
+import json
+from functools import reduce
+from pprint import pprint
+
+import pandas as pd
+from seqeval.metrics import f1_score, accuracy_score, precision_score, recall_score, classification_report
+from seqeval.metrics import performance_measure
+from seqeval.scheme import IOB2
+
+from bluesearch.mining.eval import ner_report
```
pafonta, line 28: I am a bit confused.

We end up reimplementing part of what could be found in famous packages to reduce the number of dependencies (cf. #330, #325). But here, a new dependency is introduced (i.e. `seqeval`) while there is already an implementation for this in `mining/eval.py`.

That seems confusing...

FrancescoCasalegno: Agree with @pafonta . As commented above, (1) are the metrics obtained using the two approaches the same? and (2) should we consider dropping `bluesearch.mining.eval`?

EmilieDel: Just to give you a bit of context: the output from `transformers` is a pure classification output. There is no `post-processing` on the results (which means that a token can have the class label `I-ENT` even if the previous one is not `B-ENT` or `I-ENT`). Currently, our implementation of the metrics in `bluesearch.mining.eval` requires this consistency (for the `entity` mode). As we saw `seqeval` was the library used in the scripts of the training by `transformers` (and seems to handle the inconsistency of annotations), we decided to try it and compare with our results (in the `token` mode at least). Currently, the results are not the same as the one we obtain. This is definitely something we can investigate

> should we consider dropping bluesearch.mining.eval ?

I am not sure it is needed to use `seqeval` (especially if we go for `spaCy`). However, if we want to keep it, it is important to understand some parameters (different annotations type, how they handle the inconsistencies of label, mode strict or not) and dig into their code. Moreover, it is not sure yet if and how to compute what we call the `entity` and `token` levels.

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/1_run_transformers_ner.sh
FrancescoCasalegno: Here the `$DS_VAL` is used only for the evaluation of the trained model, but no hyperparameter is chosen based on the score of the model on `$DS_VAL` (this is different from `spacy` that instead chooses the optimal `n_epoch` based on the best score achieved on the `valid_split`).

So effectively this `test` data is treated just as `test` data, not as a `valid` split, and scores are comparable.

data_and_models/pipelines/ner/transformers_vs_spacy/transformers/2_eval_pred.py
FrancescoCasalegno: Opened issue #336  to transition to `seqeval` — once we make sure that it's a good idea.

FrancescoCasalegno:
In abe7f75 we had to pin the version of `catalogue` due to an incompatibility between the requirements of `catalogue v2.0.2` and those of `twine v3.4.0`.

This change was necessary to make tests of `tox -e check-packaging` pass.

An issue was opened on `catalogue` GitHub with all details: [see here](https://github.com/explosion/catalogue/issues/17).

PR #331, opened 2021-04-02
[WIP] Change bbs_base into a multistage docker build

DriesVerachtert:
This change doesn't fix anything: it's only an attempt to reduce the size of the docker images. The original bbs_base docker image was 8.65 GB while this multistage image uses 5.72 GB instead. That's almost 3 GB less but it's still a lot and it complicates the docker image building process so don't feel obliged to actually switch to multistage builds. Let me know if this could be useful, I can also change the other images if you like.

## How to test?

It's almost not tested: I only checked if images are created and if they seem to contain the correct modules and libraries.

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/docker/base.Dockerfile
+++ b/docker/base.Dockerfile
@@ -26,7 +26,7 @@
 #
 # If the GPU support is not necessary, then another image,
 # for example "python:3.7" can be used.
-FROM nvidia/cuda:10.2-devel
+FROM nvidia/cuda:10.2-devel as builder
 
 # ARGs are only visible at build time and can be provided in
 # the docker-compose.yml file in the "args:" section or with the
@@ -35,6 +35,7 @@ ARG BBS_http_proxy
 ARG BBS_https_proxy
 ARG BBS_HTTP_PROXY
 ARG BBS_HTTPS_PROXY
+ARG BBS_USERS
 
 # ENVs are visible both at image build time and container run time.
 # We want the http proxys to be visible in both cases and therefore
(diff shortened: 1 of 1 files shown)


docker/base.Dockerfile
```diff
@@ -110,29 +91,87 @@ apt-get install -y \
...
+curl -sL https://deb.nodesource.com/setup_10.x -o nodesource_setup.sh &&\
+bash nodesource_setup.sh &&\
+rm nodesource_setup.sh &&\
+apt-get install -y nodejs; \
+DEBIAN_FRONTEND="noninteractive" \
+apt-get install -y --no-install-recommends \
+    libfontconfig1 wkhtmltopdf libmysqlclient20 \
+    python3.7-venv python3-pip && \
+python3.7 -m pip install --upgrade pip setuptools wheel && \
+update-alternatives --install /usr/local/bin/python python /usr/bin/python3.7 0; \
```
Stannislav, line 163: Hi Dries, thanks for the suggestion. This block of code appears twice. Is that intentional? And if yes, what's the ratonale?

FrancescoCasalegno: As far I understand that's because this is a new stage (see `FROM nvidia/cuda:10.2-runtime` line above).

Stannislav: Hm, maybe I misunderstand multistage builds then. My understanding was that you install a bunch of stuff in the first stage, then copy over the binaries in the second stage. But here we run pretty much the same `apt-get install` commands in both stages, this I don't understand. Or is it about installing the `*-dev` packages in the first stage and the non-dev ones in the second stage? How does this work exactly? Event then, there's overlap between the two stages, e.g. in both we do `apt-get install git htop less man vim`.

Stannislav:
Hi @DriesVerachtert 

I found time to do some experimenting. I had the suspicion that the main culprit for the bloated size were
- usage of the `nvidia/cuda:10.2-devel` base image rather than `nvidia/cuda:10.2-runtime`
- installation of dev system packages rather than non-dev ones

In view of this what I tried was to stick with the single-stage build but start with the `nvidia/cuda:10.2-runtime` base straight away and also install non-dev system packages where possible.

It turned out that this way the build is much faster and the differenc ein the resulting images is only 300MB:
```
$ docker images
bbs_base                      dries_pr_no_multistage          0083bc4358fa        18 minutes ago      6.11GB
bbs_base                      dries_pr                        1e73db2546da        57 minutes ago      5.81GB
```

Apart from that since the creation of this PR a number of things have changed in the main branch, including some dependencies. So I took the latest `Dockerfile` we had and changed it along the lines you suggested (`runtime` base + non-dev packages) and the resulting image is even smaller:
```
bbs_base                      latest                          2238124661f2        10 seconds ago      4.4GB
```

So if you don't mind I'll create a new PR with the changes based on the latest `Dockerfile` and we close this one. What do you think?

Stannislav:
Many changes suggested in this PR were successfully incorporated in the merge request #392 so that this PR can now be closed.

PR #334, opened 2021-04-08
Reduce dependencies

### LINKED ISSUE #330: Reduce number of dependencies of bluesearch
(See also #326)

In the past we discussed the importance of making optimizing the number of dependencies we need install to be able to use `bluesearch`. 

PR  #326 goes in this direction, but further actions to remove more dependencies are still possible — I would just like to discuss them before implementing them.

### 1. `joblib`
Used only to `dump()` and `load()` models from `sklearn`.
```bash
data_and_models/pipelines/sentence_embedding/train.py:81:            joblib.dump(model, f)
data_and_models/pipelines/sentence_embedding/train.py:93:            joblib.dump(model, f)
src/bluesearch/embedding_models.py:183:        self.model = joblib.load(self.checkpoint_path)
```
#### Options to remove this dependency
1. Use `pickle` instead of `joblib` — it should be possible, although maybe a bit less efficient ([source](https://scikit-learn.org/stable/modules/model_persistence.html#python-specific-serialization)).
2. Stop supporting the `sklearn` models for sentence embedding.

### 2. `scikit-learn`
In `src/` it is only for the confusion matrix, while in `data_and_models` it is used for the cosine similarity and for the 
```bash
data_and_models/pipelines/sentence_embedding/eval.py:31:from sklearn.metrics.pairwise import cosine_similarity
data_and_models/pipelines/sentence_embedding/train.py:28:from sklearn.feature_extraction.text import CountVectorizer, TfidfVectorizer
src/bluesearch/mining/eval.py:584:            data=sklearn.metrics.confusion_matrix(
```
#### Options to remove this dependency
1. If we implement the confusion matrix ourselves (doable, but will be almost a copy-paste of `sklearn`'s code), we can move `scikit-learn` from `install_requires` to `extra_requires`.
2. If we also implement the cosine distance ourselves (trivial)  and if we supporting the `sklearn` models for sentence embedding, we can remove `scikit-learn` altogether.

### 3. `tqdm`
It is only used for `bluesearch.utils.H5` for the methods `load()` and `find_unpopulated_rows()`.
```bash
src/bluesearch/utils.py:308:                iterable = tqdm.tqdm(iterable)
src/bluesearch/utils.py:410:                iterable = tqdm.tqdm(iterable)
```
#### Options to remove this dependency
1. This package is not strictly needed (if `verbose=False` it is not even used at all), so we may just remove `tqdm` and that's it.
2. If we want to provide some logging info to the user, we can write that manually e.g.  add a line `if i%100: print(f"Done {i} iterations")`.

### 4. `matplotlib`
In the `src/` part, it is only used once for plotting the confusion matrix.
```bash
src/bluesearch/mining/eval.py:649:        f, ax = plt.subplots(figsize=(len(e_types_pred) * 0.6, len(e_types_true) * 0.6))
```
#### Options to remove this dependency
1. We could generate the plot of confusion matrix directly in the `eval.py` script. This solution allows to move the dependency to `extra_requires`.
2. We could decide that we do do not want to generate such plots, a `.csv` with the confusion matrix could also be enough. This solution allows to remove the dependency altogether.

### 5. `pdfkit`
In the `src/`, this is only used used to create the PDF reports of the Search widget.
```bash
src/bluesearch/widgets/article_saver.py:301:        pdfkit.from_string(article_report, output_file_path)
src/bluesearch/widgets/search_widget.py:939:            pdfkit.from_string(
```
In fact, this Python package is just a wrapper of the library `wkhtmltopdf`, so to be able to use `pdfkit` one also needs to ensure that the binaries `wkhtmltopdf` are installed and in the current `PATH`. In practice, this point has turned out to be a big issue in the past (see e.g. experience with OOD).

#### Options to remove this dependency
1. Ask users if this feature of generating a report is still a must-have or nice-to-have. If not, drop the dependency.
2. As a compromise, we may generate a `.html` report that can be opened with any browser with a double click—sure, it's less straighforward than a PDF but maybe still fine for the users.

FrancescoCasalegno:
Fixes #330 .

### TODO
- [x] `tqdm`
- [x]  `joblib`
- [x]  `matplotlib`
- [x]   `pdfkit`
- [x]  `cryptography`

## Description

1. Removed `joblib` and `tqdm`.
2. Moved `matplotlib` from `install_requires` to `extra_requires`.
3. Reports are now in a `html` format rather than `pdf`, so `pdfkit` could be dropped.

**Note**
In #330 we said that the reason why we had `matplotlib` is in `install_requires` was to support a functionality that plots a confusion matrix that is used in `data_and_models/sentence_embedding/eval.py`.
This was in fact incorrect, the function requiring `matplotlib` was plotting another confusion matrix i.e. `plot_ner_confusion_matrix()`. But since that functionality is also not strictly needed we can drop it anyway.

## How to test?

## Checklist

  (if needed)
  (if needed)
- [x] All CI tests pass.


tests/test_widgets/test_article_saver.py
```diff
@@ -120,7 +120,7 @@ def test_summaries(self, fake_sqlalchemy_engine, tmpdir):
...
         assert isinstance(summary_table, pd.DataFrame)
 
-        # PDF report - probably can't do this on Travis...
-        # assert len(tmpdir.listdir()) == 0
-        # article_saver.pdf_report(tmpdir)
-        # assert len(tmpdir.listdir()) == 1
+        # Make report
+        assert len(tmpdir.listdir()) == 0
+        article_saver.make_report(tmpdir)
+        assert len(tmpdir.listdir()) == 1
```
EmilieDel, line 126: Really cool that we can test this now!

EmilieDel (approved):
Nice cleaning 😄 Thanks for this PR!

data_and_models/pipelines/sentence_embedding/train.py
```diff
@@ -77,8 +77,8 @@ def main():
         )
         model._stop_words_id = 0
         print("Saving model to disk...")
-        with open(out_dir / "model.pkl", "wb") as f:
-            joblib.dump(model, f)
+        with (out_dir / "model.pkl").open("wb") as f:
+            pickle.dump(model, f)
```
jankrepl, line 81: Just out of curiosity, is this significantly slower than using `joblib` or it makes no difference?

jankrepl (approved):
Looks good to me! Really nice work:))

pafonta (approved):

src/bluesearch/widgets/article_saver.py
```diff
@@ -251,6 +250,7 @@ def pdf_report(self, output_dir=None):
 
         Returns
         -------
+        output_file_path : str
```
Stannislav, line 253: Was this line duplicated by accident? Also I just noticed that the return type is actually `pathlib.Path` and not `str`.

src/bluesearch/widgets/search_widget.py
```diff
@@ -896,18 +895,18 @@ def _create_saving_checkboxes(self, article_id, paragraph_id):
...
 
-    def _cb_bt_pdf_report_article_saver(self, change_dict):
+    def _cb_bt_make_report_article_saver(self, change_dict):
         """Create the saved articles report."""
         with self.widgets["status"]:
             print()
-            print("Creating the saved results PDF report... ")
-            self.article_saver.pdf_report()
+            print("Creating the saved results report... ")
+            self.article_saver.make_report()
```
Stannislav, line 903: `make_report()` returns the file path of the saved report - we could actally print it for the user as well, something like
```python
report_path = self.article_saver.make_report()
print(f"The report was saved to {report_path.resolve()}")
```

src/bluesearch/widgets/search_widget.py
```diff
@@ -936,9 +935,8 @@ def _cb_bt_pdf_report_search(self, change_dict):
...
-                f"<style> {css_style} </style>"
-                + hyperparameters_section
-                + results_section,
-                f"report_{datetime.datetime.now()}.pdf",
-            )
+            output_file = pathlib.Path(f"report_{datetime.datetime.now()}.html")
+            with output_file.open("w") as f:
+                f.write(f"<style> {css_style} </style>")
+                f.write(hyperparameters_section)
+                f.write(results_section)
```
Stannislav, line 942: So this works without including the `<!DOCTYPE html>`, `<html>` etc. tags?

Stannislav (approved):
Pretty nice change I find. I left some minor comments here and there.

src/bluesearch/widgets/article_saver.py
FrancescoCasalegno: Thanks! Fixed in 0d9b563dd67d7b79201fae60f90394d549a2369a.

src/bluesearch/widgets/search_widget.py
FrancescoCasalegno: Good idea! Done in 2dd1414.

src/bluesearch/widgets/search_widget.py
FrancescoCasalegno: 1. `<!DOCTYPE html>` is a special beast,  and seems to be required for for legacy reasons. Modern browsers should not have issues even if it's omitted, but I think there's nothing harmful in adding it. Done in 158935e.
2. `<html>`, `<head>` ... can apparently be omitted as per [HTML specs](https://html.spec.whatwg.org/multipage/syntax.html#syntax-tag-omission).

data_and_models/pipelines/sentence_embedding/train.py
FrancescoCasalegno: I tested it and in our use case it seems that `pickle` is actually faster!

Possible explanation [here](https://stackoverflow.com/a/12617603):
> if you don't pickle large numpy arrays, then regular pickle can be significantly faster, especially on large collections of small python objects (e.g. a large dict of str objects) because the pickle module of the standard library is implemented in C while joblib is pure python.

requirements.txt
```diff
@@ -32,14 +32,12 @@ https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_craft_
 https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_jnlpba_md-0.4.0.tar.gz
```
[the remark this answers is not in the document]
FrancescoCasalegno: I started working on it but I think it's growing a bit too big to fit in the scope of this PR.

Created issue https://github.com/BlueBrain/Search/issues/344 to handle this.

Issue #335, opened 2021-04-13
Use individual spaCy with transformer backbones for all NER models


FrancescoCasalegno:
## 🚀 Feature

In light of what we discussed in PR #328, and in particular looking at the results shown in [this comparison table](https://github.com/BlueBrain/Search/pull/328#issuecomment-818738648), we should operate the following changes to the NER models in `data_and_models/pipelines/ner`.

- [x] All NER models should use a `transformer` backbone. Now they are using `tok2vec`.
- [x] All NER models should initialize the weights of this backbone using the pre-trained weights of `CORD19 NLI+STS v1`. Also, the weights should not be frozen during the fine-tuning on the NER task.
- [x] There should be one distinct NER model (= spaCy pipeline) for each entity type we support. Note that currently this is not the case, as e.g. `model2` is used to extract 3 different entity types (see table [here](https://github.com/BlueBrain/Search/blob/0b89/data_and_models/pipelines/ner/ee_models_library.csv)).
- [x] Unlike the experiments of PR #328, the spaCy pipeline should also include the rule-based entity extraction component (**Note**: for the moment let's keep using `add_er.py`, then in the future we'll improve that with #310).
- [x] All the evaluation results (token and entity based, Prec, Rec, F1) obtained **before** and **after** this PR should be collected in a table.

pafonta:
Regarding this point:

> Unlike the experiments of PR #328, the spaCy pipeline should also include the rule-based entity extraction component.

An existing issue is about the same: #310.

pafonta:
Hello @FrancescoCasalegno,

Before moving further with this task, I think some points should be clarified:

1. **Are all the checkboxes in the issue description expected to be done in this issue, which implies, in one pull request?**
Because that is a lot of work. The current issue could be split into several incremental pieces of work. As said in https://github.com/BlueBrain/Search/issues/335#issuecomment-819339941, there was for example already an issue for one point.

1. **Should this issue also deal with migrating the whole NER elements in BBS from using some models for several entities to using one model per entity?**
That is not clear at the moment which amount of extra work this task unlisted in the issue description could take (e.g. moving away from the `ee_models_library.csv` logic).

1. **Should this issue also deal with building the new train / dev sets with only one entity per set?**
Currently, 3 out of 5 annotation sets have multiple entities instead of one.

1. **Are the new models expected to use part of the `scispaCy` elements used currently? Or should it be a brand new `["transformer", "ner"]` pipeline like in #328?**
[...]

Issue #280, opened 2021-03-12
Separate source code from code for data & models


FrancescoCasalegno:
## 🚀 Feature

There are two parts in the BBS repository.
 * First part.
    * processing the source data (i.e. CORD-19),
    * training models (sentence embeddings, NERs),
    * pre-computing inference results from the models,
 * Second part.
    * defining server features (search, mining, notebooks BBS-257),
    * configuring and initializing these servers,
    * defining graphical user interfaces (notebooks, widgets).

In the context of DVC:
* the first part *generates* what DVC tracks,
* the second part *retrieves and uses* what DVC tracks.

Separating code for data+models from code for usage will make *easier to use DVC and maintain BBS*.

## Note

The goal is not to split the repository per se. Just a reorganization / cleaning could be sufficient (e.g. 3 main top-level directories: `bluesearch/`, `dvc/`, `spacy/` — or  `data/`, `models/`, `src/`).

EmilieDel:
As discussed with @jankrepl, it could be nice to:
- Start testing the code contained in the DVC part
- Create a separate folder (`experiments` or `sandbox` for example) where we can put all "random" experiments that we want to keep somewhere, without really testing and maintaining the code. (i.e. transformers vs spacy experiments, ...)

PR #345, opened 2021-04-15
Split requirements

### LINKED ISSUE #344: Refactor requirements.txt to reduce install times
## Context
Similar to what we did for  the `install_requires` dependencies in the `setup.py` (see #325), we should partition the `requirements.txt` into more than one file, to avoid long installing times when we `pip install -r requirements.txt`.

See  also https://github.com/BlueBrain/Search/pull/334#discussion_r610494187 and https://github.com/BlueBrain/Search/pull/328#discussion_r610466440 for more details.

FrancescoCasalegno:
Fixes #344 but refactoring `requirements.txt` into different files.

## What changed?
1. `bbs.Dockerfile`
   - it looks like this file was just an ancestor of `base.Dockerfile`, and now no one is using it really
   - **solution**: removed file
1. `scipy`
   - used only in `data_and_models/pipelines/sentence_embedding/eval.py`
   - was duplicated: found in `install_requires` as well as `extras_require.data_and_models`
   - **solution**: removed from `install_requires`, kept in `extras_require.data_and_models`
1. `gunicorn`
   - used in `embedding.Dockerfile`
   - used in `search.Dockerfile`
   - used in `mining.sh` — which is called by `mining.Dockerfile`
   - it's not a requirement of `bluesearch` per se (e.g. one could use `flask` instead)
   - **solution:** removed from `requirements.txt`, and added `pip install gunicorn` in `Dockerfile`s that use it.
1. `en_core_web_lg`
   - only used in `data_and_models/pipelines/ner/transformers_vs_spacy/`
   - should be added to `extra_requires.data_and_models` for consistency, but we cannot since otherwise we cannot upload to pypi
   - **solution**: keep
1. `en_core_sci_lg`
   - used only in  `src/bluesearch/database/cord_19.py` for the `CORD19DatabaseCreation`
   - no dedicated `Dockerfile` is available for the creation of the `CORD-19` database, afaik this is currently done manually!
   - **solution**: drop (if the user that manually triggers the creation of the `CORD-19` db doesn't have it, it will anyway be [informed that this model has to be installed](https://github.com/BlueBrain/Search/blob/197099cfa2636b52afd0a77557d6f45ef5c53b36/src/bluesearch/utils.py#L543))
1. `en_ner_bc5cdr_md`, `en_ner_bionlp13cg_md`, `en_ner_craft_md`, `en_ner_jnlpba_md`
    - are used as base models in the fine-tuning `data_and_models/pipelines/ner/`
    - but once we start using our version of `BioBERT` as backbone, they can all be removed (see https://github.com/BlueBrain/Search/issues/335#issuecomment-820498678)
    - **solution**: keep (for the moment)
1. `jupyterlab`
    - was listed only in `requirements.txt`, but not in `setup.py`
    - de-facto necessary to be able to use our widgets
    - **solution**: added to `install_requires` in `setup.py`
1. versions in requirements bumped up to most recent release
1. `requirements.txt` split in 3 files
   - `requirements.txt` — contains dependencies of `install_requires`
   - `requirements-dev.txt` — contains dependencies of `extras_require.dev`
   - `requirements-data_and_models.txt` — contains dependencies of `extras_require.data_and_models` plus (a) `en_ner_bc5cdr_md`, `en_ner_bionlp13cg_md`, `en_ner_craft_md`, `en_ner_jnlpba_md`; (b) `en_core_sci_lg`

deps-bef-aft

## Checklist

  (if needed)
  (if needed)
- [x] All CI tests pass.


pafonta (approved):
Thank you for doing this!

That's a nice step towards #280.

data_and_models/pipelines/sentence_embedding/Dockerfile
```diff
@@ -37,6 +37,7 @@ RUN git checkout $BBS_REVISION
 # remove ruamel-yaml: https://github.com/pypa/pip/issues/5247#issuecomment-381550610
 RUN rm -rf /opt/conda/lib/python3.8/site-packages/ruamel*
 RUN pip install -r requirements.txt
+RUN pip install -r requirements-data_and_models.txt
```
EmilieDel, line 40: Just not to forget (maybe not necessary to do it), we change the `Dockerfile`, then the corresponding `dvc.lock` are going to change `md5` of this `Dockerfile`.

EmilieDel (approved):
Thanks a lot for the splitting! LGTM!

data_and_models/pipelines/sentence_embedding/Dockerfile
FrancescoCasalegno: Yes at some point we should definitely re-run the `dvc repro` since the dependencies have indeed changed. Even if the models have not changed, the versions pinned in these `requirements*txt` files have changed, so it is very possible that we do get slightly different results.

On the time _when_ we should do this, I think that this `dvc repro` should be done when we release a new version, which is not the case yet.

EmilieDel: Sounds a good compromise to do before a release, I did not know about this! Thanks!

PR #348, opened 2021-04-26
Fix bugs in NER evaluation

### LINKED ISSUE #336: Use seqeval for NER model evaluation
## 🚀 Feature

We have seen in PR #329 that [`seqeval`](https://github.com/chakki-works/seqeval) is a good framework that implements many metrics for the evaluation of NER models.

We should avoid reinventing the wheel, so we can probably replace most of the code in `bluesearch.mining.eval` with calls to this library. But we must be careful with this, since this change may also impact the scores computed in `data_and_models/pipelines/ner/eval.py`.

## Details
- [x] Investigate whether/why the scores computed using our functions are different from those computed using `seqeval`. Assume that `y_1` and `y_2` are both valid IOB NER annotations: we should e.g. test this for the inter-rater annotations and to evaluate the predictions of our various NER models.
- [x] See if there is any functionality that we have in `bluesearch.mining.eval` but is not implemented in `seqeval`.
- [x] See if there would be any other drawback in using `seqeval`. 
- [x] Change our code to use `seqeval` instead of our custom evaluation functions.

FrancescoCasalegno:
Fixes #336.

## Description

This PR fixes bugs in our NER evaluation functions of `bluesearch.mining.eval`.
These discrepancies were first identified [here](https://github.com/BlueBrain/Search/issues/336#issuecomment-826830138) and [here](https://github.com/BlueBrain/Search/issues/336#issuecomment-823244116).

## How to test?

Unit tests were added to introduce cases that used to fail and are now passing.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
- [x] All CI tests pass. 
- [x] Add more docs on `IOB2` and meaning of `TP`, `FN`, `FP` errors.
- [x] Add informative error when `pred` or `true` is inconsistent + add utils to convert to `IOB2`
- [x] Add type annotations.

### DIFF
--- a/src/bluesearch/mining/eval.py
+++ b/src/bluesearch/mining/eval.py
@@ -374,9 +374,9 @@ def idx2text(tokens, idxs):
 def ner_report(
     iob_true: pd.Series,
     iob_pred: pd.Series,
-    mode: Optional[str] = "entity",
+    mode: str = "entity",
     etypes_map: Optional[dict] = None,
-    return_dict: Optional[bool] = False,
+    return_dict: bool = False,
 ) -> Union[str, OrderedDict]:
     """Build a summary report showing the main ner evaluation metrics.
 
@@ -488,9 +488,9 @@ def ner_errors(
     iob_true: pd.Series,
     iob_pred: pd.Series,
     tokens: pd.Series,
-    mode: Optional[str] = "entity",
+    mode: str = "entity",
     etypes_map: Optional[dict] = None,
-    return_dict: Optional[bool] = False,
+    return_dict: bool = False,
 ) -> Union[str, OrderedDict]:
     """Build a summary report for the named entity recognition.
 
@@ -599,7 +599,7 @@ def ner_confusion_matrix(
     iob_true: pd.Series,
     iob_pred: pd.Series,
     normalize: Optional[str] = None,
-    mode: Optional[str] = "entity",
+    mode: str = "entity",
 ) -> pd.DataFrame:
     """Compute confusion matrix to evaluate the accuracy of a NER model.
 
@@ -612,7 +612,7 @@ def ner_confusion_matrix(
     iob_pred :
         Predicted IOB2 annotations.
     normalize :
-        One of "true", "pred", "all".
+        One of "true", "pred", "all", or None.
         Normalizes confusion matrix over the true (rows), predicted (columns)
         conditions or all the population. If None, the confusion matrix will
         not be normalized.
--- a/src/bluesearch/mining/eval.py
+++ b/src/bluesearch/mining/eval.py
@@ -22,7 +22,6 @@ import json
 import string
 from collections import OrderedDict
 from pathlib import Path
-from typing import Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


jankrepl:
Looks great!
I run the script from the linked issue ([link](https://github.com/BlueBrain/Search/issues/336#issuecomment-823244116)) and it seems like now `bluesearch_entity == seqeval_IOB2` for all etypes.

eval_bug_fix

src/bluesearch/mining/eval.py
```diff
@@ -423,34 +486,40 @@ def ner_report(iob_true, iob_pred, mode="entity", etypes_map=None, return_dict=F
 
 
 def ner_errors(
-    iob_true, iob_pred, tokens, mode="entity", etypes_map=None, return_dict=False
-):
+    iob_true: pd.Series,
+    iob_pred: pd.Series,
+    tokens: pd.Series,
+    mode: Optional[str] = "entity",
```
jankrepl, line 492: I might be wrong here, but I think `Optional[str]` is equivalent to `Union[str, None]`. IMO it means that the user can pass a `None` and the code will handle it accordingly (it does not mean "this is a keyword argument that has a default"). However, the code cannot handle `mode=None` and would raise an exception.

See https://mypy.readthedocs.io/en/stable/getting_started.html#the-typing-module for references.

I feel like the meaning is different from the `numpydoc`'s `optional`.

src/bluesearch/mining/eval.py
```diff
@@ -423,34 +486,40 @@ def ner_report(iob_true, iob_pred, mode="entity", etypes_map=None, return_dict=F
...
+    etypes_map: Optional[dict] = None,
+    return_dict: Optional[bool] = False,
```
jankrepl, line 494: Same comment as above. I mean in `if` statements `None` will behave like a `False` and it passing `return_dict=None` is not going to error out but I guess the same argument  as above applies.

src/bluesearch/mining/eval.py
```diff
@@ -523,20 +596,28 @@ def ner_errors(
         return "\n".join(out)
 
 
-def ner_confusion_matrix(iob_true, iob_pred, normalize=None, mode="entity"):
+def ner_confusion_matrix(
+    iob_true: pd.Series,
+    iob_pred: pd.Series,
+    normalize: Optional[str] = None,
+    mode: Optional[str] = "entity",
```
jankrepl, line 603: Same as above

jankrepl (commented):
Great PR and nice find!  Thank you for the work. See some minor comments in the code

src/bluesearch/mining/eval.py
```diff
@@ -31,7 +31,66 @@
 from spacy.tokens import Doc
 
 
-# TODO : remove references to
+def _check_consistent_iob(iob_true: pd.Series, iob_pred: pd.Series) -> None:
```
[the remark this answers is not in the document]
FrancescoCasalegno: I think it makes sense to keep this function private. It's a utility meant for internal usage, I don't think users would really benefit from being able to call it.

Would you prefer to make it public (i.e. remove initial `_`)? Or at least tell `Sphinx` to include this function?
Or is it OK as it is now for you as well?

src/bluesearch/mining/eval.py
FrancescoCasalegno: Very interesting, I totally missed this and I indeed thought it was like `numpydoc`'s `optional!

I think this also applies to this line (even if it's not from this PR):
https://github.com/BlueBrain/Search/blob/b9c886966ca4d893b41457a17262e198e3ba7f03/src/bluesearch/embedding_models.py#L303

Fixed all wrong usages of `Optiona[...]` in 1ecd77b.

src/bluesearch/mining/eval.py
FrancescoCasalegno: Fixed all wrong usages of Optiona[...] in 1ecd77b.

src/bluesearch/mining/eval.py
FrancescoCasalegno: Fixed all wrong usages of Optiona[...] in 1ecd77b.

src/bluesearch/mining/eval.py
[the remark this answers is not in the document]
jankrepl: IMO we can leave it the way it is now:)

jankrepl (approved):

EmilieDel (approved):
Thanks for this PR! Really nice finding, it is great that our results are now equivalent to `seqeval`! Maybe it could be nice to force the recomputation of the metrics under `/data_and_models/`, what do you think ?

tests/test_mining/test_eval.py
```diff
@@ -578,3 +601,38 @@ def test_ner_errors(ner_annotations, dataset, mode, errors_expected):
...
     pd.testing.assert_frame_equal(df_after, punctuation_annotations["after"])
+
+
+@pytest.mark.parametrize(
+    "iob_pred, raises",
+    [
+        (pd.Series(["O", "B-a", "B-a", "I-a", "B-c", "O"]), False),
+        (
+            pd.Series(["O", "B-a", "B-a", "I-a", "B-c"]),
+            "inconsistent numbers",
```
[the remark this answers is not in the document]
FrancescoCasalegno: Changed to a more explicit `"target variables with inconsistent numbers of samples"` in 687a048.

Also notice that this string we test for must be a substring of the exception message thrown here
https://github.com/BlueBrain/Search/blob/687a04841170e61d1883ed76c3fb13fda4f6bf6c/src/bluesearch/mining/eval.py#L56-L60
which I wrote by taking inspiration from what `scikit-learn` does (see [here](https://github.com/scikit-learn/scikit-learn/blob/c957eb37b5988e6e2a4692c1356e8689294404c5/sklearn/utils/validation.py#L305-L320)).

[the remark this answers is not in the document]
EmilieDel: I see! Thanks for the explanation!

PR #350, opened 2021-04-28
#347 - #327 Upgrade spaCy>=3.0.6

### LINKED ISSUE #327: Upgrade to the spaCy with the patch for a deterministic vocab/strings.json
## Version

https://github.com/explosion/spaCy/pull/7603 is included in [spaCy v3.0.6](https://github.com/explosion/spaCy/compare/v3.0.5...v3.0.6).

## Context

During #248, it was discovered that the order of lines in `model-best/vocab/strings.json` is non-deterministic.

A patch was applied in #248 to make it deterministic.

A PR has been submitted to spaCy for doing the same upstream: https://github.com/explosion/spaCy/pull/7603.

## Goal

1. Remove the patch file `data_and_models/pipelines/ner/patch.py` and the calls in `dvc.yaml`.

2. Put as minimal version in `setup.py` and `requirements.txt` the spaCy version containing https://github.com/explosion/spaCy/pull/7603.

3. Re-run `dvc repro -f` in `data_and_models/pipelines/ner/`.

4. Check that there is no change after another `dvc repro -f`.

5. Commit `dvc.lock` and execute `dvc push`.

### LINKED ISSUE #347: Upgrade to the spaCy with a built-in copy_vocab_tokenizer()
## Version

https://github.com/explosion/spaCy/pull/7750 is included in [spaCy v3.0.6](https://github.com/explosion/spaCy/compare/v3.0.5...v3.0.6).

## Context

During #248, it was discovered that the custom tokenizer and vocabulary from `scispaCy` models aren't copied by default.

A custom solution was applied in #248 to copy these tokenizer and vocabulary.

The feature has been implemented upstream in PR https://github.com/explosion/spaCy/pull/7750.

## Goal

1. Replace in `data_and_models/pipelines/ner/config.cfg`
```
[initialize.before_init]
@callbacks = "copy_vocab_tokenizer"
source = ${paths.base_model}
```
by
```
[initialize.before_init]
@callbacks = "spacy.copy_from_base_model.v1"
tokenizer = ${paths.base_model}
vocab = ${paths.base_model}
```

2. Remove the custom solution file `data_and_models/pipelines/ner/initialize.py` and the references in `dvc.yaml`.

3. Put as minimal version in `setup.py` and `requirements.txt` the spaCy version containing https://github.com/explosion/spaCy/pull/7750.

4. Re-run `dvc repro -f` in `data_and_models/pipelines/ner/`.

5. Check that there is no change after another `dvc repro -f`.

6. Commit `dvc.lock` and execute `dvc push`.

EmilieDel:
Fixes #347.
Fixes #327.

## Description

From #347:
> During #248, it was discovered that the custom tokenizer and vocabulary from scispaCy models aren't copied by default.
> A custom solution was applied in #248 to copy these tokenizer and vocabulary.
> The feature has been implemented upstream in PR explosion/spaCy#7750.

From #327:
> During #248, it was discovered that the order of lines in model-best/vocab/strings.json is non-deterministic.
> A patch was applied in #248 to make it deterministic.
> A PR has been submitted to spaCy for doing the same upstream: explosion/spaCy#7603.

Done during the PR:

1. [#347] Replaced the `[initialize.before_init]` part in `data_and_models/pipelines/ner/config.cfg`.
2. [#347] Removed file `data_and_models/pipelines/ner/initialize.py` and the references in `dvc.yaml`.
3. [#327] Removed file `data_and_models/pipelines/ner/patch.py` and the calls in `dvc.yaml`.
4. [Both] Put as minimal version in `setup.py` and `requirements.txt` the `spaCy` version handling #7750 and #7603 (`spaCy >= 3.0.6`).

## How to test?

```
dvc pull
dvc repro
```
should not have any changes. 

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


setup.py
```diff
@@ -66,8 +66,8 @@
     # >= 0.4.0 to be compatible with spaCy 3.
     "scispacy>=0.4.0",
     "sentence-transformers",
-    # >= 3.0.4 to include the fix for https://github.com/explosion/spaCy/issues/7352.
-    "spacy>=3.0.4",
+    # >= 3.0.6 to include the fix for https://github.com/explosion/spaCy/issues/7750.
```
pafonta, line 69:
```suggestion
    # >= 3.0.6 to include the fix for https://github.com/explosion/spaCy/pull/7603.
```

Indeed:
1. 7603 is the fix while 7750 is a new feature,
2. 7750 shouldn't be needed if we use raw spaCy pipelines (decision from Tuesday in the context of #335),
3. so I would suggest to have 7603 in the comment, as it seems to stay the most relevant in the long term.

pafonta (changes requested):
Thank you for taking care of this!

I have a small suggested change that could be merged right away.

Otherwise, I have tested with the following and there was no change (as expected):
```
dvc repro -f evaluation@disease
```

pafonta (approved):
Thank you for applying the requested changes :)

jankrepl (approved):
Perfect! Thanks!

data_and_models/pipelines/ner/dvc.lock
```diff
@@ -12,12 +12,12 @@ stages:
...
-      md5: 12bfb6ab6750bb670005b4722ff05104.dir
-      size: 171634265
+      md5: 189c70c7e75a4c9c83c986d03460cc23.dir
+      size: 171634305
       nfiles: 21
     - path: Dockerfile
-      md5: 9a4e8a0fffb0a7e49faa347f2ec2132d
-      size: 1732
+      md5: 7863c45504f96d4d5147ae07f8918759
+      size: 1772
```
FrancescoCasalegno, line 20: 1. It looks like all the  `ner/modelX/model-best/` and all the `ner/modelX/model-best/` have changed. Do you have an idea why?
2. On the contrary, none of the `../../metrics/ner/xxx.json` seems to have changed. How does this relates to the previous point? By the way, if some metric had changed the `xxx.json` should have also been `git add`ed to this PR right?
3.  The `Dockerfile` was not touched in this PR, is it normal that its `md5` has changed?

FrancescoCasalegno (commented):
Thanks for your work, definitely simplifies a lot the process!

I have left few questions on the `dvc.lock`, other than that 👍  from my side.

data_and_models/pipelines/ner/dvc.lock
EmilieDel: 1. Yes it is. Each `ner/modelX/model-best/` folders contains a copy of the `config.cfg`. From #347, we needed to change the initialization directly in that file. This is the reason why the `md5` changed. In terms of implementation, there is no change in the logic (you can compare the [spacy commit](https://github.com/explosion/spaCy/commit/bdb485cc80ca6822471eac0a29ac2782d9b55450) and our previous [initialize.py](https://github.com/BlueBrain/Search/blob/master/data_and_models/pipelines/ner/initialize.py) )
2. The `md5` of the model has changed but the weights did not. Side note concerning this, we recently changed the implementation of our `eval.py` but the version of BBS pinned in the `Dockerfile` is still `v0.1.2`. Once this pinned version is going to change, the `metrics/ner/xxx.json` are going to be affected. Those files are tracked with `git`, it is going to be needed to be add them once it is done.
3. `Dockerfile` was touched in [previous commit](https://github.com/BlueBrain/Search/commit/b9c886966ca4d893b41457a17262e198e3ba7f03) but the `md5` was not updated at that time. Personally, I don't think it is a problem we don't update `dvc.lock` each time but I can understand it is confusing!

FrancescoCasalegno (approved):

PR #359, opened 2021-05-05
Patch torch.save() to ensure deterministic behavior

### LINKED ISSUE #343: Test reproducibility of spaCy training
## Context
We have already seen (see [BBS-198](https://bbpteam.epfl.ch/project/issues/browse/BBS-198)) that `torch` training results  (i.e. model weights) are not bitwise reproducible.

On the other hand, `spaCy` training has been reproducible until now. But we have been using `tok2vec` as a backbone, not `transformer`, so can this change the situation?

## Actions
- [x] Test if `spacy` NER training is reproducible when using a `transformer` backbone instead of `tok2vec`.
- [x] If No to the previous question, is the training also not reproducible with a frozen `transformer`.
- [x] If training appears not to be reproducible, ask `spaCy` developers if this is expected—it seems indeed to contrast with the multiple times that "reproducibile" [is mentioned in their docs](https://spacy.io/usage/training).

FrancescoCasalegno:
Fixes #343.

## Description

We have observed that `torch.save()` produces non-deterministic binary outputs. This PR introduces a patch to fix that.

The idea of this patch is to simply map the non-deterministic `_cdata` attributes to a deterministic value using a dictionary that is incrementally populated. For more details, see section `📌 Content of this PR` in https://github.com/pytorch/pytorch/pull/57536.

**Note**. This PR was also opened on `pytorch` GitHub https://github.com/pytorch/pytorch/pull/57536, but while waiting for that PR to be merged and a new version of `torch` to be released, we need something on our side to be able to move on (especially with #351).

## How to test?

Wrap your calls to `torch.save()` in the following way to ensure deterministic outputs.
```python
from unittest.mock import patch

with patch("torch.serialization._save", patched_torch_save):
    torch.save(obj, out_file)
```

See also examples in section `🧪 How to test` in https://github.com/pytorch/pytorch/pull/57536.

## Checklist

- [x] Unit tests added.
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/utils.py
+++ b/src/bluesearch/utils.py
@@ -17,15 +17,18 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+import io
 import json
 import pathlib
 import time
-from typing import Any, Set, Union
+from typing import Any, Dict, Set, Union
 
 import h5py
 import numpy as np
 import pandas as pd
 import spacy
+import torch
+from torch.serialization import location_tag, normalize_storage_type
 
 
 class Timer:
@@ -575,3 +578,54 @@ If `model_name` is a package name, please install it using
 If `model_name` is a local path, please verify the pipeline path.
 """
             ) from err
+
+
+def patched_torch_save(obj, zip_file, pickle_module, pickle_protocol):
+    """Patches torch.serialization._save() to have deterministic binary output.
+
+    See related PR [1] and original source code for torch.serialization._save()
+    [2].
+
+    References
+    ----------
+    [1] https://github.com/pytorch/pytorch/pull/57536
+    [2] https://github.com/pytorch/pytorch/blob/133d8abbfc764a6e32e54a44113f46e715fe385c/torch/serialization.py#L457-L498
+    """  # noqa
+    serialized_storages = {}
+    id_map: Dict[int, str] = {}
+
+    def persistent_id(obj):
+        # FIXME: the docs say that persistent_id should only return a string
+        # but torch store returns tuples. This works only in the binary protocol
+        # see
+        # https://docs.python.org/2/library/pickle.html#pickling-and-unpickling-external-objects
+        # https://github.com/python/cpython/blob/master/Lib/pickle.py#L527-L537
+        if torch.is_storage(obj):
+            storage_type = normalize_storage_type(type(obj))
+            obj_key = id_map.setdefault(obj._cdata, str(len(id_map)))
+            location = location_tag(obj)
+            serialized_storages[obj_key] = obj
+
+            return ("storage", storage_type, obj_key, location, obj.size())
+        return None
+
+    # Write the pickle data for `obj`
+    data_buf = io.BytesIO()
+    pickler = pickle_module.Pickler(data_buf, protocol=pickle_protocol)
+    pickler.persistent_id = persistent_id
+    pickler.dump(obj)
+    data_value = data_buf.getvalue()
+    zip_file.write_record("data.pkl", data_value, len(data_value))
+
+    # Write each tensor to a file named tensor/the_tensor_key in the zip archive
+    for key in sorted(serialized_storages.keys()):
+        name = f"data/{key}"
+        storage = serialized_storages[key]
+        # given that we copy things around anyway, we might use storage.cpu()
+        # this means to that to get tensors serialized, you need to implement
+        # .cpu() on the underlying Storage
+        if storage.device.type != "cpu":
+            storage = storage.cpu()
+        # Now that it is on the CPU we can directly copy it into the zip file
+        num_bytes = storage.size() * storage.element_size()
+        zip_file.write_record(name, storage.data_ptr(), num_bytes)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,15 +17,26 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+import hashlib
 import pathlib
+from unittest.mock import patch
 
 import h5py
 import numpy as np
 import pandas as pd
 import pytest
 import spacy
+import torch
+from torch.nn import Linear, Sequential
 
-from bluesearch.utils import H5, JSONL, Timer, load_ee_models_library, load_spacy_model
+from bluesearch.utils import (
+    H5,
+    JSONL,
+    Timer,
+    load_ee_models_library,
+    load_spacy_model,
+    patched_torch_save,
+)
 
 
 class TestTimer:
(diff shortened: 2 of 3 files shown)


src/bluesearch/utils.py
```diff
@@ -575,3 +578,54 @@ def load_spacy_model(
...
+        name = f"data/{key}"
+        storage = serialized_storages[key]
+        # given that we copy things around anyway, we might use storage.cpu()
+        # this means to that to get tensors serialized, you need to implement
+        # .cpu() on the underlying Storage
+        if storage.device.type != "cpu":
+            storage = storage.cpu()
+        # Now that it is on the CPU we can directly copy it into the zip file
+        num_bytes = storage.size() * storage.element_size()
+        zip_file.write_record(name, storage.data_ptr(), num_bytes)
```
jankrepl, line 631: I wonder how this `storage.data_ptr()` is used because similarily to `_cdata` it will be session specific

tests/test_utils.py
```diff
@@ -377,3 +388,34 @@ def test_load_spacy_model(model_name, is_found):
...
+    ],
+)
+def test_patched_torch_save(tmpdir, get_obj):
+    md5_sums = []
+
+    for _ in range(2):
+        torch.manual_seed(42)
+        obj = get_obj()
+
+        file_out = pathlib.Path(str(tmpdir)) / "output.pt"
```
jankrepl, line 412: I guess this one can be moved outside of the `for` loop

tests/test_utils.py
```diff
@@ -377,3 +388,34 @@ def test_load_spacy_model(model_name, is_found):
...
+        },
+        lambda: Sequential(Linear(3, 5), Linear(5, 2)),
+    ],
+)
+def test_patched_torch_save(tmpdir, get_obj):
+    md5_sums = []
+
+    for _ in range(2):
+        torch.manual_seed(42)
+        obj = get_obj()
```
jankrepl, line 410: I guess by using `lambda` you are trying to defer instantiation to this line, right? Was it necessary? I guess just become of the possibility to control random state?

tests/test_utils.py
```diff
@@ -377,3 +388,34 @@ def test_load_spacy_model(model_name, is_found):
...
+    "get_obj",
+    [
+        lambda: torch.tensor([8.0, 8.0, 5.0]),
+        lambda: {
+            "a": torch.tensor([8.0, 8.0, 5.0]),
+            "b": torch.tensor([3.0, 1.0, 2.0, 4.0]),
+            "c": torch.tensor([[1.0, 1.9], [1.2, 1.3]]),
+        },
+        lambda: Sequential(Linear(3, 5), Linear(5, 2)),
+    ],
```
jankrepl, line 403: I would add one more test case - state dictionary of a module. It is the recommended way how to save models for inference (according to https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-for-inference)

So something like 
```python
lambda: Sequential(Linear(3, 5), Linear(5, 2)).state_dict()
```
It seems to use `OrderedDict`, `tuple` and `list`

```python
>>> Sequential(Linear(3, 5), Linear(5, 2).state_dict()

OrderedDict([('0.weight',
              tensor([[ 0.4512, -0.1306, -0.0298],
                      [ 0.2635,  0.5573, -0.0610],
                      [-0.5151, -0.2372, -0.2422],
                      [-0.1536, -0.5279, -0.2040],
                      [-0.3668,  0.4431,  0.3355]])),
             ('0.bias', tensor([-0.0857,  0.3729,  0.0929,  0.3936, -0.1156])),
             ('1.weight',
              tensor([[-0.2642,  0.0674,  0.2640,  0.2075, -0.4029],
                      [-0.3660, -0.0844,  0.1445,  0.4002, -0.3518]])),
             ('1.bias', tensor([ 0.1969, -0.0726]))])
```

jankrepl (commented):
Very nice and really important!!! Great job!

See comments in the code.

tests/test_utils.py
FrancescoCasalegno: So here we have to distinguish line-by-line.

### First line
```python
        torch.manual_seed(42)
```
This really cannot go outside the `for` loop, or the `nn.Module` layers will be instantiated with different weights at each iteration, and therefore the output files will differ.

### Second line
```python
        obj = get_obj()
```
If we move this guy _outside_ the `for` loop (together with the previous one) then even w/o the patch the tests would pass! Indeed, calling `torch.save()` twice in a row on the same object has always produced the same binary output—the problem before was with objects that had identical fields but were instantiated at different times (in a way, similar to the difference between `==` and `is`).

### Third line
```python
        file_out = pathlib.Path(str(tmpdir)) / "output.pt"
```
Yes, you're right on this one! Thanks a lot, fixed in 42df3c2

tests/test_utils.py
FrancescoCasalegno: Yes, indeed, the problem was that I couldn't find another way to do that.

I wanted 2 `obj` to be created at each iteration, but with the same `seed` so that they should have equal content but different identity, `_cdata` etc.

tests/test_utils.py
jankrepl: Sorry, I was only referring to the path! That was my mistake;)

tests/test_utils.py
FrancescoCasalegno: Good idea, added in 7dfb1fa.

Anyway, notice that having a `Dict[str, torch.Tensor]` was already tested, so here the only difference is that we add a test for `OrderedDict` instead of `Dict`.

---

> It seems to use `OrderedDict`, `tuple` and `list`

As a side note, I think that what is `repr`esented as 2-ples `(key, value)` is just the way `OrderedDict` behaves, but I am not sure there are really `tuples` stored inside as `keys` or `values`.

tests/test_utils.py
FrancescoCasalegno: No problem! Feel free to "resolve" if you're satisfied then :)

tests/test_utils.py
jankrepl: Good point!

src/bluesearch/utils.py
FrancescoCasalegno: Interesting question.

Just curious, if you run examples in section `🧪 How to test` in the `pytorch` PR, do you get the exact same results (`md5` sums included) that I reported there?

jankrepl: Let me check!

jankrepl: Unfortunately,  I am getting reproducible results on my machine but the MD5 hashes are different than yours.

FrancescoCasalegno: I think there was something wrong in the results of then `🧪 How to test`...
Fixed the `md5` results on the PR on `pytorch`, now everything should be fine?

EmilieDel (approved):
Thanks a lot for this PR! It is going to help, for sure. 

Works well on my side!

FrancescoCasalegno:
### Update 2021-05-17
- `torch` merged the PR we opened on their code base to fix reproducibility: https://github.com/pytorch/pytorch/pull/57536.
- The latest release as of now was on 2021-03-25 so `pip install --ugprade torch` won't install the reproducibility patch, and we may still have to wait a bit for a new `torch` version with the patch to be released.

FrancescoCasalegno:
## To DO
- [ ] Investigate apparent non-reproducibility `torch.rand()` (is it `float32` or what??)

FrancescoCasalegno:
## Closing
`torch` people merged pytorch/pytorch#57536 and the new version `1.9.0` will be soon released — the release candidate for `1.9.0` was indeed published [here](https://github.com/pytorch/pytorch/releases) on 2021-05-18.

PR #362, opened 2021-05-07
Fix NER eval bug — dtype conversion from NumPy's to Python's

### LINKED ISSUE #361: DVC eval crashes "int64 not JSON serializable"
The DVC evaluation is crashing. After investigation, the bug was introduced by #348.

The bug:
```
  File "eval.py", line 111, in <module>
    main()
  File "eval.py", line 107, in main
    json.dump(all_metrics_dict, f)
  File "/opt/conda/lib/python3.8/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/opt/conda/lib/python3.8/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/opt/conda/lib/python3.8/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/opt/conda/lib/python3.8/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/opt/conda/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type int64 is not JSON serializable
```

To reproduce:

```
# For the bug introduced by #348, use 0bb500551b1b7c6f5bb9228335aa4df30a654e9c.
# For the working code __before__ #348, use b9c886966ca4d893b41457a17262e198e3ba7f03.
export COMMIT=...

git clone https://github.com/BlueBrain/Search
cd Search/

# Change <image> and <container>.
docker build -f data_and_models/pipelines/ner/Dockerfile --build-arg BBS_REVISION=$COMMIT -t <image> .
docker run -it --rm -v /raid:/raid --name <container> <image>

git checkout $COMMIT
git checkout -- data_and_models/pipelines/ner/dvc.lock

cd data_and_models/pipelines/ner/
dvc pull --with-deps evaluation@organism
dvc repro -fs evaluation@organism
```

_Originally posted by @pafonta in https://github.com/BlueBrain/Search/issues/335#issuecomment-833506692_

FrancescoCasalegno:
Fixes #361.

## Description

The isssue was due to the fact that the returned values of `ner_report` were the result of a reduce operation (`.sum()` in this case) on a `pd.Series` of integers. And integers are represented by `np.int64` in `pandas`, rather than the native `int` type of Python.

Two possible solutions (see also [here](https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types)).
1. Manually cast to desired type, e.g. `int(series.sum())`
2. Extract the underlying value as a Python scalar with [`.item()`](https://pandas.pydata.org/docs/reference/api/pandas.Series.item.html), e.g. `series.sum().item()`

I think option 2 is better, since it's more generic: `np.float32` will be automatically converted to `float`, `np.int64` to `int` etc. So I used this approach.

## How to test?

The test example given in #361 should now run without error.

## Checklist

- [ ] All CI tests pass.

### DIFF
--- a/src/bluesearch/mining/eval.py
+++ b/src/bluesearch/mining/eval.py
@@ -306,10 +306,10 @@ def unique_etypes(iob, return_counts=False, mode="entity"):
         return unique
     else:
         if mode == "entity":
-            unique_counts = [(iob == f"B-{etype}").sum() for etype in unique]
+            unique_counts = [(iob == f"B-{etype}").sum().item() for etype in unique]
         elif mode == "token":
             unique_counts = [
-                (iob.isin([f"B-{etype}", f"I-{etype}"])).sum() for etype in unique
+                (iob.isin([f"B-{etype}", f"I-{etype}"])).sum().item() for etype in unique
             ]
         else:
             raise ValueError(f"Mode '{mode}' is not available.")
@@ -443,9 +443,9 @@ def ner_report(
             ent_pred = iob_pred.isin(
                 [f"B-{etypes_map[etype]}", f"I-{etypes_map[etype]}"]
             )
-            n_true = ent_true.sum()
-            n_pred = ent_pred.sum()
-            true_pos = (ent_true & ent_pred).sum()
+            n_true = ent_true.sum().item()
+            n_pred = ent_pred.sum().item()
+            true_pos = (ent_true & ent_pred).sum().item()
         else:
             raise ValueError(f"Mode {mode} is not available.")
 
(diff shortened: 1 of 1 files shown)


pafonta (approved):
Nice, thank you!

jankrepl (approved):
Makes sense!

Stannislav (approved):
Yes, looks good to me.

Unrelated to this PR, but probably instead of things like
```python
iob_true.isin([f"B-{etype}", f"I-{etype}"])
```
the following would work as well
```python
iob_true.str.fullmatch(fr"(B|I)-{etype}")
```
Not sure which one's faster though.

PR #364, opened 2021-05-19
Add new linting components

### LINKED ISSUE #363: Add new linting components
# New Flake8 Components

We could include the following components in our linting pipeline in tox:

- `flake8-docstrings` (replaces `pydocstyle`)
- `flake8-bugbear` (find code that’s likely to be a bug)
- `flake8-comprehensions` (write list/set/dict comprehensions in the recommended form)

Replacing `pydocstyle` by the`flake8-docstrings` wrapper would give use the benefit of using the superior `flake8` CLI and configuration system. We would also be able to run all checks with one `flake8` command.

# Improve Configuration
Instead of passing various permanent options as CLI commands we should add these option to configuration files. This way they would also apply when running the CLI tools like `flake8` and `isort` by hand int he command line.

Stannislav:
Closes #363 .

## Description

- Add `flake8-bugbear`, `flake8-comprehensions`, and `flake8-docstrings` to the linting pipeline
- Add `isort` configuration to `tox.ini` (instead of providing configuration on the command line)
- Upgrade `black` and `isort`
- Upgrade `pandas-stubs`, this makes the artificial casting to `pd.DataFrame` that we introduced obsolete.
- Streamline the command line parameters in the `lint` and `format` tox sections
- Add possibility to provide files for linting and formatting as positional arguments to tox.

## How to test?

The main changes are in `tox.ini`, all other changes are fixes for new linting errors.

The linting should pass:
```bash
$ tox -e lint
```

Furthermore, now it's also possible to lint/format selected paths by providing them as positional arguments to tox:
```bash
$ tox -e format -- custom_path
$ tox -e lint -- custom_path
```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass. 

# Follow-Up Ideas
Should we lint/test the `.py` files in `data_and_models` as well? Right now there's complete anarchy.

### DIFF
--- a/src/bluesearch/database/mining_cache.py
+++ b/src/bluesearch/database/mining_cache.py
@@ -307,7 +307,7 @@ class CreateMiningCache:
 
     def _delete_rows(self):
         """Delete rows in the target table that will be re-populated."""
-        for model_id, model_schema in self.model_schemas.items():
+        for model_id, _model_schema in self.model_schemas.items():
             # Reformatted due to this bandit bug in python3.8:
             # https://github.com/PyCQA/bandit/issues/658
             query = (  # nosec
@@ -576,7 +576,7 @@ class CreateMiningCache:
         schema_df = self.ee_models_library
 
         model_schemas: Dict[str, Dict[str, Any]] = {}
-        for i, row in schema_df.iterrows():
+        for _i, row in schema_df.iterrows():
             entity_type_to = row["entity_type"]
             model_path = row["model_path"]
             model_id = row["model_id"]
--- a/src/bluesearch/embedding_models.py
+++ b/src/bluesearch/embedding_models.py
@@ -272,8 +272,8 @@ def compute_database_embeddings(connection, model, indices, batch_size=10):
     sentences = retrieve_sentences_from_sentence_ids(indices, connection)
     n_sentences = len(sentences)
 
-    all_embeddings = list()
-    all_ids = list()
+    all_embeddings = []
+    all_ids = []
 
     for batch_ix in range((n_sentences // batch_size) + 1):
         start_ix = batch_ix * batch_size
--- a/src/bluesearch/mining/entity.py
+++ b/src/bluesearch/mining/entity.py
@@ -19,7 +19,6 @@
 
 import ast
 import copy
-from typing import cast
 
 import numpy as np
 import pandas as pd
@@ -180,14 +179,7 @@ class PatternCreator:
             If ``int`` then represent a row index to be dropped. If ``list`` then
             a collection of row indices to be dropped.
         """
-        # Both DataFrame.drop() and DataFrame.reset_index() return None if
-        # the parameter inplace=True is set. We have inplace=False but need to
-        # cast the returned object from Optional[DataFrame] to DataFrame in
-        # order to satisfy mypy.
-        # There is a big discussion on deprecating the inplace parameter here:
-        # https://github.com/pandas-dev/pandas/issues/16529
-        self._storage = cast(pd.DataFrame, self._storage.drop(index=labels))
-        self._storage = cast(pd.DataFrame, self._storage.reset_index(drop=True))
+        self._storage = self._storage.drop(index=labels).reset_index(drop=True)
 
     def to_df(self):
         """Convert to a pd.DataFrame.
--- a/src/bluesearch/mining/eval.py
+++ b/src/bluesearch/mining/eval.py
@@ -22,7 +22,7 @@ import json
 import string
 from collections import OrderedDict
 from pathlib import Path
-from typing import Optional, Union, cast
+from typing import Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -78,8 +78,7 @@ def _check_consistent_iob(iob_true: pd.Series, iob_pred: pd.Series) -> None:
             )
 
         etypes = unique_etypes(x)
-        # cast() needed to tell mypy that fillna() doesn't return None here.
-        x_prev = cast(pd.Series, x.shift(periods=1).fillna("O", inplace=False))
+        x_prev = x.shift(periods=1).fillna("O", inplace=False)
         for etype in etypes:
             if (
                 x.isin([f"I-{etype}"]) & ~x_prev.isin([f"B-{etype}", f"I-{etype}"])
--- a/tests/test_database/test_cord_19.py
+++ b/tests/test_database/test_cord_19.py
@@ -99,7 +99,7 @@ class TestDatabaseCreation:
     def test_database_content(self, real_sqlalchemy_engine):
         """Tests that the two tables expected has been created. """
         inspector = sqlalchemy.inspect(real_sqlalchemy_engine)
-        tables_names = [table_name for table_name in inspector.get_table_names()]
+        tables_names = list(inspector.get_table_names())
         assert "sentences" in tables_names
         assert "articles" in tables_names
 
@@ -228,12 +228,8 @@ class TestDatabaseCreation:
         Tests that the schema of the fake database is always the same as
         the real one.
         """
-        real_tables_names = {
-            table_name for table_name in real_sqlalchemy_engine.table_names()
-        }
-        fake_tables_names = {
-            table_name for table_name in fake_sqlalchemy_engine.table_names()
-        }
+        real_tables_names = set(real_sqlalchemy_engine.table_names())
+        fake_tables_names = set(fake_sqlalchemy_engine.table_names())
 
         assert real_tables_names == {"articles", "sentences"}
         assert real_tables_names.issubset(fake_tables_names)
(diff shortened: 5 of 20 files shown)


src/bluesearch/database/mining_cache.py
```diff
@@ -307,7 +307,7 @@ def construct(self):
 
     def _delete_rows(self):
         """Delete rows in the target table that will be re-populated."""
-        for model_id, model_schema in self.model_schemas.items():
+        for model_id, _model_schema in self.model_schemas.items():
```
jankrepl, line 310: Wow, why?

src/bluesearch/embedding_models.py
```diff
@@ -272,8 +272,8 @@ def compute_database_embeddings(connection, model, indices, batch_size=10):
     sentences = retrieve_sentences_from_sentence_ids(indices, connection)
     n_sentences = len(sentences)
 
-    all_embeddings = list()
-    all_ids = list()
+    all_embeddings = []
+    all_ids = []
```
jankrepl, line 276: What is the reasoning here? It seems like this is a reoccurring patter: `[]` instead of `list()`, `{}` instead of `dict()`,...

src/bluesearch/database/mining_cache.py
EmilieDel: I am interested to know too :) Is it because it is not used in the `for` loop ? If yes, maybe we could just replace by `_` instead ?

src/bluesearch/mining/eval.py
```diff
@@ -78,8 +78,7 @@ def _check_consistent_iob(iob_true: pd.Series, iob_pred: pd.Series) -> None:
             )
 
         etypes = unique_etypes(x)
-        # cast() needed to tell mypy that fillna() doesn't return None here.
-        x_prev = cast(pd.Series, x.shift(periods=1).fillna("O", inplace=False))
+        x_prev = x.shift(periods=1).fillna("O", inplace=False)
```
EmilieDel, line 81: It is nice that we don't need to use `cast` anymore! It is thanks to the upgrade of the `pandas_stubs` ?

EmilieDel (approved):

src/bluesearch/mining/eval.py
Stannislav: Yes I think so, once I upgraded `pandas_stubs` it started saying the cast was unnecessary.

src/bluesearch/database/mining_cache.py
Stannislav: Yes, exactly, it's because this variable isn't used in the loop body. This was reported by `flake8-bugbear`. I think it's legit - it could be that you're looping over something but due to a typo use a different variable than that you're looping over.

I don't mind using `_` either, to me the only difference might be the readability that the underscored name still let's you understand what kind of objects are being iterated over and why they're not used. But otherwise I don't really mind. What do you guys prefer?

src/bluesearch/embedding_models.py
Stannislav: Apparenty slightly faster and slightly safer as per [here](https://stackoverflow.com/a/6610783/2804645) and [here](https://softwareengineering.stackexchange.com/a/319549/109896).

To be honest, I'm not completely sold on the speed and safety gains. Probably best to view it as a convention to be used throughout the codebase.

src/bluesearch/database/mining_cache.py
EmilieDel: Keeping track of the variable name is not a bad idea! But I don't have a strong opinion about this!

jankrepl: I am surprised that this is not caught by vanilla `flake8` - after all it is an unused variable, no?

src/bluesearch/mining/entity.py
```diff
@@ -180,14 +179,7 @@ def drop(self, labels):
...
         """
-        # Both DataFrame.drop() and DataFrame.reset_index() return None if
-        # the parameter inplace=True is set. We have inplace=False but need to
-        # cast the returned object from Optional[DataFrame] to DataFrame in
-        # order to satisfy mypy.
-        # There is a big discussion on deprecating the inplace parameter here:
-        # https://github.com/pandas-dev/pandas/issues/16529
-        self._storage = cast(pd.DataFrame, self._storage.drop(index=labels))
-        self._storage = cast(pd.DataFrame, self._storage.reset_index(drop=True))
+        self._storage = self._storage.drop(index=labels).reset_index(drop=True)
```
pafonta, line 182: Nice that the double cast is no more needed!

tox.ini
```diff
@@ -141,6 +135,11 @@ show_missing = False
 count = False
 max-line-length = 88
 extend-ignore = E203
+docstring-convention = numpy
+per-file-ignores =
+    tests/*:D
+    benchmarks/*:D
```
pafonta, line 141: Is it introducing a new behaviour? Or is it a migration from something from before?

pafonta (commented):
Thank you for adding these tools to check more best practices!

src/bluesearch/database/mining_cache.py
Stannislav: that's true, don't know why flake8 doens't spot it by itself

tests/test_database/test_cord_19.py
```diff
@@ -97,9 +97,9 @@ class TestDatabaseCreation:
     """Tests the creation of the Database"""
 
     def test_database_content(self, real_sqlalchemy_engine):
-        """Tests that the two tables expected has been created. """
+        """Tests that the two tables expected has been created."""
         inspector = sqlalchemy.inspect(real_sqlalchemy_engine)
-        tables_names = [table_name for table_name in inspector.get_table_names()]
+        tables_names = list(inspector.get_table_names())
```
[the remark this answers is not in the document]
Stannislav: I think it was `flake8-comprehensions`

tox.ini
```diff
@@ -39,31 +33,31 @@ basepython = python3.7
...
-    isort --honor-noqa --profile black --check setup.py {[tox]source} tests benchmarks
-    pydocstyle {[tox]source}
-    black -q --check setup.py {[tox]source} tests benchmarks
-    bandit -qr {[tox]source}
-    mypy --config-file .mypy.ini  setup.py {[tox]source} tests
-    mypy --config-file .mypy.ini  {[tox]source} benchmarks
+    flake8 {posargs:{[tox]sources}}
+    isort --check {posargs:{[tox]sources}}
+    black --check {posargs:{[tox]sources}}
+    bandit -qr --exclude "benchmarks/,tests/" {posargs:{[tox]sources}}
```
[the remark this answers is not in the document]
Stannislav: I thought about it, but I'm a bit hesitant. When everything passes, black only prints 2 additional lines:
```text
lint run-test: commands[2] | black --check setup.py src/bluesearch tests benchmarks
All done! ✨ 🍰 ✨
68 files would be left unchanged.
```
but bandit spams the stdout with a huge chunk of text, even if there's not issues:
```text
lint run-test: commands[3] | bandit -r --exclude benchmarks/,tests/ setup.py src/bluesearch tests benchmarks
[main]  INFO    profile include tests: None
[main]  INFO    profile exclude tests: None
[main]  INFO    cli include tests: None
[main]  INFO    cli exclude tests: None
[main]  INFO    running on Python 3.7.9
[node_visitor]  INFO    Unable to find qualified name for module: setup.py
Run started:2021-05-21 10:00:23.943386

Test results:
        No issues identified.

Code scanned:
        Total lines of code: 7646
        Total lines skipped (#nosec): 7

Run metrics:
        Total issues (by severity):
                Undefined: 0.0
                Low: 0.0
                Medium: 0.0
                High: 0.0
        Total issues (by confidence):
                Undefined: 0.0
                Low: 0.0
                Medium: 0.0
                High: 0.0
Files skipped (0):
```
so I decided to keep it silenced, especially because we don't really have to fix bandit-related stuff very often.

What do you think?

tox.ini
```diff
@@ -141,6 +135,11 @@ show_missing = False
...
+docstring-convention = numpy
+per-file-ignores =
+    tests/*:D
+    benchmarks/*:D
 
-[pydocstyle]
-convention = numpy
+[isort]
+profile = black
+honor_noqa = true
```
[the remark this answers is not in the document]
Stannislav: Yep, now one can also run `isort` in the terminal by itself too without having to type these additional options.

tox.ini
Stannislav: It's migration of the old behaviour, but slightly disguised. What these options do is disabling `pydocstyle`-related warnings (`DXXX`) for tests and benchmarks, since we don't write docstrings for them.

Previously we just didn't run `pydocstyle` on these two folders, but now we have to because it's part of `flake8`.

So all in all, this reproduces the old behaviour.

tox.ini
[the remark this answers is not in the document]
pafonta: I see. I am OK with the current implementation then.

pafonta (approved):

PR #388, opened 2021-06-17
Fix RTD build fail

### LINKED ISSUE #387: RTD doc is failing
## 🐛 Bug description

The docs badge on the README shows that RTD fails to build the documentation.

## To reproduce

One could have a look at the build logs here:
https://readthedocs.org/projects/blue-brain-search/builds/

## Expected behavior

The docs badge on the README should show that RTD succeeds to build the documentation.

## Environment

The master branch on GitHub.

## Additional context

It seems that the cause is from #301.

The PR removed `mysqlclient` in `INSTALL_REQUIRES` but forgot to remove these lines:
https://github.com/BlueBrain/Search/blob/de86792d5260c78ced8f3ade35282e132aa391cc/setup.py#L74-L76

Stannislav:
Fixes #387.

## Description

See the issue #387 

## How to test?

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ INSTALL_REQUIRES = [
     # >= 0.4.0 to be compatible with spaCy 3.
     "scispacy>=0.4.0",
     "sentence-transformers",
+    "setuptools>=45",
     # >= 3.0.6 to include the fix for https://github.com/explosion/spaCy/pull/7603.
     "spacy[transformers]>=3.0.6",
     "torch",
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,6 @@ INSTALL_REQUIRES = [
     # >= 0.4.0 to be compatible with spaCy 3.
     "scispacy>=0.4.0",
     "sentence-transformers",
-    "setuptools>=45",
     # >= 3.0.6 to include the fix for https://github.com/explosion/spaCy/pull/7603.
     "spacy[transformers]>=3.0.6",
     "torch",
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,6 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 """The setup script."""
-import os
-
 from setuptools import find_packages, setup
 
 DESCRIPTION = (
@@ -70,11 +68,6 @@ INSTALL_REQUIRES = [
     "spacy[transformers]>=3.0.6",
     "torch",
 ]
-
-if os.environ.get("READTHEDOCS") == "True":
-    # see https://github.com/readthedocs/readthedocs-docker-images/issues/158
-    INSTALL_REQUIRES.remove("mysqlclient")
-
 
 EXTRAS_REQUIRE = {
     "dev": [
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,7 +7,7 @@ sphinx:
   configuration: docs/conf.py
 
 build:
-    image: 6.0
+    image: "6.0"
 
 python:
   version: 3.7
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,7 +7,7 @@ sphinx:
   configuration: docs/conf.py
 
 build:
-    image: stable
+    image: latest
 
 python:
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -66,16 +64,12 @@
     # >= 0.4.0 to be compatible with spaCy 3.
     "scispacy>=0.4.0",
     "sentence-transformers",
+    "setuptools>=45",
```
Stannislav, line 67: I'm curious what lead to this requirement?

Would it also be possible to move this to `pyproject.toml` because I imging this is only required at install time but not at run time.

FrancescoCasalegno:
## Fixing `setuptools` version issues
The job on RTD kept failling with the following error:
```python
ERROR: Command errored out with exit status 1:
   command: /home/docs/checkouts/readthedocs.org/user_builds/blue-brain-search/envs/issues-387/bin/python /home/docs/checkouts/readthedocs.org/user_builds/blue-brain-search/envs/issues-387/lib/python3.7/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /tmp/tmpdn0x70a1
...
  pkg_resources.VersionConflict: 
(
    setuptools 41.0.1 (/home/docs/.pyenv/versions/3.7.9/lib/python3.7/site-packages), 
    Requirement.parse('setuptools>=45')
)
```

## What I tried and did _not_ work
RTD automatically upgrades `setuptools`, and the version seems compatible than the required `>= 45`
```python
/home/docs/checkouts/readthedocs.org/user_builds/blue-brain-search/envs/issues-387/bin/python -m pip install --upgrade --no-cache-dir pip setuptools
...
Requirement already satisfied: setuptools in /home/docs/checkouts/readthedocs.org/user_builds/blue-brain-search/envs/issues-387/lib/python3.7/site-packages (57.0.0)
```
... yet we had the error shown above.

So I tried to force to upgrade `setuptools` version by setting a min version in the `setup.py` or in the `pyproject.toml`. 
Still, I kept having this weird error about a mysterious old version of our package, this `setuptools 41.0.1`.

## What I tried and _did_ work

So the origin of this old version of `setuptools` that RTD insists in using finally seems to come from the Docker image used by RTD. Now, we were using `image: stable`, which corresponds (see [here](https://github.com/readthedocs/readthedocs-docker-images#read-the-docs---docker-images)) to the image tagged as `5.0`. And this version [has the line](https://github.com/readthedocs/readthedocs-docker-images/blob/718820847d1d48a3e9624ba03761ea018757c0fc/Dockerfile#L180)
```dockerfile
ENV RTD_SETUPTOOLS_VERSION 41.0.1
```
Setting `image: "6.0"` lets RTD use a version [which has finally a compatible version of `setuptools`](https://github.com/readthedocs/readthedocs-docker-images/blob/10b47cee232449b5d84f0e066b97281d76982308/Dockerfile#L22)
```dockerfile
ENV RTD_SETUPTOOLS_VERSION 45.1.0
```
And that indeed fixes the issue.

## Warning
The info on the available image tags on RTD's own docs (see [here](https://kodok.readthedocs.io/en/stable/yaml-config.html#build-image)) appear to be outdated, as they don't mention any image version after `2.0`, while on DockerHub [you can see](https://hub.docker.com/r/readthedocs/build/) that there are versions up to `7.0` and the current `stable` is in fact `5.0`.

setup.py
FrancescoCasalegno: Yes, clearly this `setuptools` didn't belong here, I removed it in a following commit.

I was mainly trying things around to fix an RTD error related to inconsitent `setuptools` version.

I added more details in my comment [here](https://github.com/BlueBrain/Search/pull/388#issuecomment-869686029).

Stannislav:
I guess the quotes around "6.0" are necessary because othewise it's interpreted as a float while a string is expected?

Stannislav:
I'm not sure I actually understand if `setuptools>=45` is a requirement that *we* need or *RTD* needs.

Otherwise, thanks for fixing :)

EmilieDel (approved):
Thanks for the comment explaining the changes and thanks for the fix!

PR #389, opened 2021-06-25
Remove unused SciSpacy models and obsolete mentions thereof

### LINKED ISSUE #381: Remove scispaCy and its unused models
## Context

As of v0.2.0a (i.e. from #351):
- we do `not` use any more the package `scispaCy`,
- we use `only` its model `en_core_sci_lg` (see next line).

```
$ grep -ir --include "*.py" --include "*.txt" "en_core_sci_" .
./src/bluesearch/database/cord_19.py:    def _sentences_table(self, model_name="en_core_sci_lg"):
```

⚠️ While we use the model `en_core_sci_lg`, we `never install it`.

## Goal

1 - Add  the model `en_core_sci_lg` to our dependencies.

2 - Remove `scispaCy` and all other models from `dependencies` and `comments`.

The following commands should return `nothing`:
```
grep -ir --include "*.py" --include "*.txt" "scispacy" .
```
```
grep -ir --include "*.py" --include "*.txt" "en_ner_" .
```

Stannislav:
Fixes #381 .

## Description

## How to test?

```bash
git grep -in "scispacy"
```
```bash
git grep -in "en_ner"
```
```bash
git grep -in "en_core_sci"
```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [ ] All CI tests pass.


Stannislav:
Occurences of SciSpacy to keep
==============================
en_ner_*
---------
- `data_and_models/annotations/ner/README.md`: These are descriptions
   about how the annotations were collected. We want to keep this historical record.
- `notebooks/NER_Evaluation.ipynb`: still unclear what to do with it

en_core_sci_lg
---------------
- We're still using it in two places:
  - `bluesearch.database.cord_19::CORD19DatabaseCreation`
  - `notebooks/STS_evaluation.ipynb`
- Therefore the corresponding dependency is in `requirements-data_and_models.txt`

SciSpacy
---------
- `data_and_models/annotations/ner/patterns/README.md`: just an example, can be left there
- `docs/source/whatsnew.rst`: part of the history
- `notebooks/NER_Evaluation.ipynb`: same as for `en_ner_*`
- `requirements-data_and_models.txt`: same as for `en_core_sci_lg`
- `setup.py` and `requirements.txt`: install `scispacy==0.4.0`, I suppose we still need to install the
   `scispacy` package as long as we're using `en_core_sci_lg`?

Stannislav:
The notebook `NER_Evaluation.ipynb` just got deleted.

I think this PR is done and ready for review @pafonta @FrancescoCasalegno

pafonta (changes requested):
`scispacy` needs to be deleted as a dependency.

> `setup.py` and `requirements.txt`: install `scispacy==0.4.0`, I suppose we still need to install the`scispacy` package as long as we're using `en_core_sci_lg`?

Not that I know.

EmilieDel (approved):
Regarding having `scispacy` in `setup.py` and `requirements.txt`, it seems you can download and run the model without `scispacy` installed. However, they obviously recommend to install `scispacy` first before downloading any `scispacy` model. Personally, I don't mind keeping it as we use one of their model in the code anyway!

data_and_models/pipelines/ner/transformers_vs_spacy/spacy/eval.py
```diff
@@ -16,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
-
+"""Evaluation script for NER models."""
```
FrancescoCasalegno, line 17: Ditto — I would keep the docstring above.

notebooks/demo_attribute_extraction.ipynb
```diff
@@ -53,7 +53,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "entity_model = spacy.load(\"en_ner_craft_md\")"
+    "entity_model = spacy.load(\"data_and_models/models/ner_er/model-chemical\")"
```
FrancescoCasalegno, line 56: What is this `ner_er/model-chemical`?
I couldn't find it tracked with `dvc` and I think I have not really seen it before...

requirements-data_and_models.txt
```diff
@@ -18,10 +18,7 @@
 PyYAML==5.4.1
 dvc[ssh]==2.0.17
 https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_bc5cdr_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_bionlp13cg_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_craft_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_jnlpba_md-0.4.0.tar.gz
+https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_lg-0.4.0.tar.gz
```
FrancescoCasalegno, line 21: As you indeed say in the PR description, the only place where this `en_core_sci_lg` is used — other than the `STS_Evaluation` notebook — is for `bluesearch.database.cord_19::CORD19DatabaseCreation`.

So if has to be in the requirements, shouldn't it be in the main `requirements.txt` rather than in the `data_and_models` one? The latter is — as far as I understand — mainly for the dependencies of what we have in the dvc stuff inside `data_and_models/`.

FrancescoCasalegno (commented):
Awesome, thanks a lot for the amazing cleanup!  🧹 ✨ 

As for whether to keep `scispacy` in the `setup.py` and in the `requirements.txt`, I am a bit confused.
1. Nowhere in the `src/` code we are now **assuming** that `scispacy` is installed, so I think it is safe to remove it from the `setup.py`.

2. As @EmilieDel [said](https://github.com/BlueBrain/Search/pull/389#pullrequestreview-694836972) I also remember being able to install  and run a `scispacy` model w/o need for installing `scispsacy` before that. However, [their docs say](https://github.com/allenai/scispacy#installation) that (i) `scispacy` module defines custom pipes (inlcuding custom tokenizers and (ii) using scispacy "requires two steps: installing the library and intalling the models". So there are two possibilities now.
    - If running `pip install scispacy` does impacts (maybe because of the usage of custom pipes like the tokenizer?) on the output of the `scipspacy` models, then I think we should keep it in the `requirements.txt`, but still remove from `setup.py`.
    - If that is not the case, we can remove it also from `requirement.txt`.

data_and_models/pipelines/ner/add_er.py
```diff
@@ -16,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
-
+"""Append an entity ruler to a spacy pipeline."""
```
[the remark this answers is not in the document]
Stannislav: Oh I thought it was common to have the license as the top block of a source file and then everything else, so I though this was a typo and corrected it. Do we want to revert it?

By the way, doc-checker completely ignores the license block and correctly recognises the module docstring as the first line of the source file.

data_and_models/pipelines/ner/eval.py
```diff
@@ -16,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
-
+"""Evaluation script for NER models."""
```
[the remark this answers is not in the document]
Stannislav: see above

data_and_models/pipelines/ner/transformers_vs_spacy/spacy/eval.py
Stannislav: see above

notebooks/demo_attribute_extraction.ipynb
Stannislav: It's listed in the `data_and_models/models/ner_er/.gitignore` as well as in `data_and_models/pipelines/ner/params.yaml` so I assumed it's a legit model that we have trained.

EmilieDel: Sorry for jumping in out of the blue, `ner_er/model-XX` (XX being `entity type`) are the new models trained after:
- switching to `transformers` backbone 
- the decision of one model per entity type was taken.

Stannislav:
@EmilieDel @FrancescoCasalegno @pafonta thanks for testing the `scispacy` dependency, I completely removed it now in 
d2f01d8

requirements-data_and_models.txt
Stannislav: Good point, both `sci-lg` and `web-sm` models are now in `requirements.txt` as per commit d42ce01.

The particular syntax, including the dashes, is how the corresponding entries appear in `pip freeze`.

notebooks/demo_attribute_extraction.ipynb
FrancescoCasalegno: Ah thanks @EmilieDel! I had an older branch checked-out locally and couldn't see it.

data_and_models/pipelines/ner/add_er.py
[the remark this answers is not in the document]
FrancescoCasalegno: It's the same for me, I just thought it would be good to be consistent.
Any option is fine with me.

FrancescoCasalegno (approved):

requirements-data_and_models.txt
pafonta: Actually, for `en_core_web_sm`, shouldn't it be in `requirements-dev.txt`? It is used only in the tests, no?

Stannislav: Right. It's now in `requirements-dev.txt`.

requirements-data_and_models.txt
```diff
@@ -14,14 +14,8 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
-
```
pafonta, line 17: This line break shouldn't go away, isn't it?

Stannislav: you want the empty line back?

pafonta: This is what we have in the other two requirements files. So, it will introduce an inconsistency.

Stannislav: Fixed in 5f633d3

data_and_models/pipelines/ner/add_er.py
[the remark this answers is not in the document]
Stannislav: I made everything consistent with the rest in f028969

PR #390, opened 2021-06-25
Add missing sklearn.metrics import in eval module

Stannislav:
## Description

In `bluesearch.minig.eval.ner_confusion_matrix` we use `sklearn.metrics.confusion_matrix`
but the corresponding import was missing.

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/src/bluesearch/mining/eval.py
+++ b/src/bluesearch/mining/eval.py
@@ -27,6 +27,7 @@ from typing import Optional, Union
 import numpy as np
 import pandas as pd
 import sklearn
+import sklearn.metrics
 from spacy.tokens import Doc
 
 


EmilieDel (approved):

FrancescoCasalegno:
Our tests always passed, including this one
https://github.com/BlueBrain/Search/blob/0bb500551b1b7c6f5bb9228335aa4df30a654e9c/tests/test_mining/test_eval.py#L497
So is this `import` necessary?

Notice that the submodule `metrics` is part of the `__all__` defined in `sklearn`'s `__init__.py`:
https://github.com/scikit-learn/scikit-learn/blob/8e8e6102b461ac91c7e8aa4ce55acfea512e8bbc/sklearn/__init__.py#L106

Stannislav:
@FrancescoCasalegno @pafonta 

I think we don't get an error because `sklearn.metrics` is probably imported by some other module that we import and that's how `sklearn.metrics` ends up in `sys.modules`. But I don't think relying on some module to import another module is a good idea.

I honestly have no idea what's going on in `sklearn/__init__.py` - they list `"metrics"` in their `__all__` but I don't see it being imported anywhere. To my knowledge all that's listed in `__all__` has to be imported first.

Here's a small experiment I just ran to convince myself that importing `sklearn` does not make `sklearn.metrics` available out of the box:

```python
# main.py
import my_module
import sklearn

print(sklearn.metrics.confusion_matrix)
```

```python
# my_module.py
# empty for the moment`
```

```bash
$ python main.py
 main.py
  File "main.py", line 5, in <module>
    print(sklearn.metrics.confusion_matrix)
AttributeError: module 'sklearn' has no attribute 'metrics'
```

Now change `my_module.py` to
```python
# my_module.py
import sklearn.metrics
```
and get
```bash
$ python main.py
<function confusion_matrix at 0x11c969280>
```
which shows that importing `sklearn.metrics` in a module that you import also makes it available in the current module.

But like I said, I find relying on a module being imported somewhere by someone is not really safe and the most straightforward way is to just import all modules explicitly that are used in the code.

FrancescoCasalegno (approved):

pafonta:
@Stannislav

Not being able to use `sklearn.metrics.confusion_matrix(...)` after `import sklearn` seems to be an error from `scikit-learn`.

Indeed, many other libraries let do such thing (see below).

> But I don't think relying on some module to import another module is a good idea.
> ...
> But like I said, I find relying on a module being imported somewhere by someone is not really safe and the most straightforward way is to just import all modules explicitly that are used in the code.

Why not. But then, we should be consistent and fix everywhere where we do this. For example:
here where `sqlalchemy.sql` is not imported:
https://github.com/BlueBrain/Search/blob/12988ef564dd4e6373a7455f5ee30c0608e2e972/src/bluesearch/database/cord_19.py#L336
and here where `np.random` is not imported:
https://github.com/BlueBrain/Search/blob/12988ef564dd4e6373a7455f5ee30c0608e2e972/tests/test_embedding_models.py#L314
and here where `pytest.mark` is not imported:
https://github.com/BlueBrain/Search/blob/12988ef564dd4e6373a7455f5ee30c0608e2e972/tests/test_mining/test_eval.py#L43

Stannislav:
@pafonta I don't know why you're trying to make this simple and clear issue unnecessarily complicated, the time wasted on this is incredible

- `sklearn.metrics.confusion_matrix` doesn't work without importing
- if a package pulls one of it's submodules up the hierarchy by importing it in its `__init__.py` that's fine (cf. `pytest.mark`, `np.random` etc.)
- this is not happening with `sklearn.metrics`
- conting on a different package to import `sklearn.metrics` is a bad idea

what else is there to discuss? It's a one line change that fixes something that is broken.

FrancescoCasalegno:
Hi @Stannislav and @pafonta,

I think that the need for this 1-line change has been well clarified.

1. For whatever reason `sklearn` doesn't `import` its sub-modules, unlike what done by `numpy` [here](https://github.com/numpy/numpy/blob/e69faef03428b82f5651a5fb521610ddea2bd22a/numpy/__init__.py#L158-L163), hence the need to `import sklearn.metrics`.
2. Things were working in our code before even w/o this fix, but that was just by accident and we cannot rely on side-effects of other `import`s in general.

Let's merge this PR and move on with other issues! 🙂

PR #392, opened 2021-06-28
Improve docker base image

Stannislav:
## Description

This is a follow-up to https://github.com/BlueBrain/Search/pull/331

The main reasons for the reduction of the `bbs-base` image size seem to be
- using the `-runtime` version of the CUDA base image instead of `-devel`
- installing non-dev system packages
- installation of Node.js is not necessary any more

Since Dries' PR some dependencies were removed from `bluesearch` as well, which might have contributed to the size reduction too.

Current image sizes:
```
bbs-base                      latest                          509489f9476d        2 days ago          5.27GB
bbs_base                      v4                              510b086603e5        3 months ago        8.65GB
```
Other changes:
- Rename docker images `bbs_base` to `bbs-base` (I though it looks better / is easier to type, is that OK?)
- Fix DVC NER model pulling in mining server (was actually broken since we changed the format of `dvc.yaml` and the way we do NER in general) see `docker/utils.sh::dvc_pull_models`
### Building Images
```bash
docker-compose build base
docker build -f docker/search.Dockerfile -t bbs-search .
docker build -f docker/mining.Dockerfile -t bbs-mining .
```

### Testing bbs_base
```bash
cd <bluesearch_source_dir>
docker run --rm -itu $USER -v $PWD:/workdir bbs-base
pip install --use-feature=in-tree-build ".[dev]"
pytest
```

### Testing Servers
```bash
docker run --rm -itv /raid:/raid -p 9850:8080 --env-file .env --name bbs-search-pr bbs-search
docker run --rm -itv /raid:/raid -p 9852:8080 --env-file .env --name bbs-mining-pr bbs-mining
curl -X POST dgx1.bbp.epfl.ch:9850/help | jq .
curl -X POST dgx1.bbp.epfl.ch:9852/help | jq .
# <try using search/mining in the search-graph-examples notebook>
```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


docker/utils.sh
```diff
@@ -48,8 +48,8 @@ dvc_pull_models() {
   fi
 
   # Pull the NER models in DVC
-  pushd "$DATA_AND_MODELS/pipelines/ner/" || exit
-  grep -Eo '\badd_er_[0-9]+\b' dvc.yaml | xargs dvc pull
+  pushd "$DATA_AND_MODELS/models/ner_er/" || exit
+  grep -E '^/model-' .gitignore | tr -d '/' | xargs dvc pull
```
EmilieDel, line 52: Ah nice! Good finding! It reduces the number of models downloaded by 2, right ?

requirements-data_and_models.txt
```diff
@@ -18,10 +18,7 @@
 PyYAML==5.4.1
 dvc[ssh]==2.0.17
 https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_bc5cdr_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_bionlp13cg_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_craft_md-0.4.0.tar.gz
-https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_ner_jnlpba_md-0.4.0.tar.gz
+https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_lg-0.4.0.tar.gz
```
EmilieDel, line 21: Why do we need this one ? Is it because we use it for testing purposes ?

EmilieDel (approved):
Nice work! Thanks for doing this. 

Related to `docker`, I was wondering if it can be useful to create some `.dockerignore` files mirroring the `.gitignore` ones generated from DVC pipelines. (e.g. `data_and_models/models/ner/.gitignore`, ...). Indeed, if we have some models locally pulled and we want to build an image, all those models are increasing the size of the docker image.

docker/base.Dockerfile
```diff
@@ -127,13 +127,13 @@ COPY docker/utils.sh /tmp
 RUN \
 . /tmp/utils.sh && \
 groupadd -g 999 docker && \
-create_users "${BBS_USERS},bbs_user/1000" "docker" && \
+create_users "${BBS_USERS},guest/1000" "docker" && \
```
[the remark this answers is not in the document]
Stannislav: There was a typo here anyway (creating user `bbs_user` but trying to use `bbsuser` further down) so while correcting it I thought `bbs` is not actually 100% correct any more since we're `bluesearch` and `bs_user` sounds rude, so I ended up just taking `guest`.

docker/utils.sh
Stannislav: Well actually the old grep doesn't work anymore because we're now using `foreach` and the actual model names are listed in `params.yaml`. I could've also grepped `params.yaml`, but going for `.gitignore` seemed a little bit more straighforward.

requirements-data_and_models.txt
Stannislav: It was to anticipate #389 and to already have `en-core-sci-lg` installed on the `bbs-base` image before #389 gets merged. But now I should actually move it to `requirements.txt` since that's where it's ended up in #389

Stannislav: I just actually moved it to make sure it's the same as in #389 (71c42a3)

Stannislav:
> Nice work! Thanks for doing this.
> 
> Related to `docker`, I was wondering if it can be useful to create some `.dockerignore` files mirroring the `.gitignore` ones generated from DVC pipelines. (e.g. `data_and_models/models/ner/.gitignore`, ...). Indeed, if we have some models locally pulled and we want to build an image, all those models are increasing the size of the docker image.

Yes, I also constantly run into the issue of `data_and_models` being huge and getting in the way of `docker build`. So far I've been manually deleting all DVC tracked data before building... I was wondering if there's a DVC command to do that, but wasn't able to find one.

I have nothing against `.dockerignore` mirroring all `.gitignore`s except that because it needs to be done manually they will necessarily get out of sync at some point.

PR #393, opened 2021-06-28
#375  Test linting of "data_and_models/" 

### LINKED ISSUE #375: Test linting of "data_and_models/"
Extend `lint` tests to the content of `data_and_models/`.

Any file that are taken from other repos _as-is_ (e.g. [this script](https://github.com/BlueBrain/Search/blob/f0384001c0d6dca164a159187b8d3bc4ebb839bd/data_and_models/pipelines/sentence_embedding/scripts/fine_tune.py#L1)) should not be covered by this test.

EmilieDel:
Fixes #375.

## Description

This PR handles the linting test of `data_and_models` part of the repository. 
- Folders `data_and_models/pipelines/sentence_embedding/scripts` and `data_and_models/pipelines/ner/transformers_vs_spacy/transformers/` are ignored because scripts are coming from other repositories.
- Packages of linting part are upgraded.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
- [X] `setup.py` and `requirements.txt` updated with new dependencies.
  (if needed)
- [X] Type annotations added.
- [x] All CI tests pass.


data_and_models/annotations/ner/analyze.py
```diff
@@ -75,31 +75,34 @@
 
 
 def report(input_path: Path, verbose: bool) -> List[Union[int, float]]:
+    """Create report of the annotations."""
     print(f"Analyzing {input_path}...")
 
-    results = []
+    results: List[float] = []
```
pafonta, line 81:
```suggestion
    results: List[Union[int, float]] = []
```

`results` is what is returned by the function. The return type of the function should then be used.

pafonta (commented):
Thanks for working on #375!

I have just made a small comment on a wrong typing.

Otherwise, there are changes to a lot of different parts of our code. I think as much as possible people should have a look.

data_and_models/pipelines/ner/dvc.yaml
```diff
@@ -88,14 +88,14 @@ stages:
     foreach: ${entities}
     do:
       cmd: >-
-        python eval.py
+        python eval_ner.py
```
Stannislav, line 91: Is this because of mypy complaining about modules with the same names?

data_and_models/pipelines/relation_extraction/convert_chemprot_fmt.py
```diff
@@ -251,7 +288,17 @@ def annotate_scibert(row):
...
+            part_1
+            + "<< "
+            + entity_1
+            + " >>"
+            + part_2
+            + "[[ "
+            + entity_2
+            + " ]]"
+            + part_3
+        )
```
Stannislav, line 301: Maybe this one can be made into an f-string?

src/bluesearch/mining/eval.py
```diff
@@ -362,6 +363,29 @@ def idx2text(tokens, idxs):
...
+    iob_true: pd.Series,
+    iob_pred: pd.Series,
+    mode: str = "entity",
+    etypes_map: Optional[dict] = None,
+    *,
+    return_dict: Literal[True],
+) -> OrderedDict:
+    ...
+
+
```
Stannislav, line 388: Hm, interesting - what's the background for this and what does it do?

tox.ini
```diff
@@ -33,21 +33,28 @@ basepython = python3.7
...
 commands =
     flake8 {posargs:{[tox]sources}}
     isort --check {posargs:{[tox]sources}}
     black --check {posargs:{[tox]sources}}
-    bandit -qr --exclude "benchmarks/,tests/" {posargs:{[tox]sources}}
-    mypy --exclude "benchmarks/conftest.py" {posargs:{[tox]sources}}
+    bandit -qr --exclude "benchmarks/,tests/,data_and_models/pipelines/ner/transformers_vs_spacy/transformers/" \
+    {posargs:{[tox]sources}}
+    mypy \
+    --exclude "benchmarks/conftest.py|data_and_models/pipelines/ner/transformers_vs_spacy/transformers/|data_and_models/pipelines/sentence_embedding/scripts/" \
```
Stannislav, line 56: I think the exclude configuration can also be put into the `.mypy.ini` file. Is there anything that speaks against it?

I think the advantage would be that the setting would also apply when you run `mypy` from the command line.

data_and_models/pipelines/ner/dvc.yaml
EmilieDel: Indeed, there are several `eval.py` into `data_and_models` folder. `Mypy` is complaining about it.

src/bluesearch/mining/eval.py
EmilieDel: In several `eval.py` from `data_and_models` folder, we call this function `ner_report` and we go through the output `dictionary` thanks to `.items()`. However, from typing, `mypy` does not know if the output is a `str` or a `dict` and thus is complaining that `str` does not have any `items()` method. 

One solution for this, it is to use `cast()` (as we did in the past for `pandas`). However, in this specific case, we know from `return_dict` value what is going to be the output type. This part of the code is specifying that if `return_dict` is `True`, the output is going to be a `dict`, otherwise, it is going to be a `str`. Doing that, you do not need to cast your function anymore.

Stannislav: Ah I see, thanks for the explanation!

To me this looks like a good reason for not using variable return types :)

EmilieDel: I think it is inspired by `sklearn.confusion_matrix`. However, I am against changing this part of the code to avoid variable return types. I don't think we have a lot of methods in this case except this one.

tox.ini
EmilieDel: Done in 99b63f4.

data_and_models/pipelines/relation_extraction/convert_chemprot_fmt.py
EmilieDel: Done in 99b63f4.

src/bluesearch/mining/eval.py
Stannislav: No, no, of course not, I'm not suggesting we should refactor our code, that would clearly be outside of the scope of this PR. It was just a general observation :)

FrancescoCasalegno:
## Note

Let's please wait for #394 to be merged before merging this PR into `master` :)

Stannislav (approved):

Issue #355, opened 2021-05-04
Investigate reproducibility of Sentence Embedding training with Patch


FrancescoCasalegno:
Our Sentence Embdedding training scripts (see [here](https://github.com/BlueBrain/Search/tree/master/data_and_models/pipelines/sentence_embedding/scripts)) are under `data_and_models/` but their output is not tracked with DVC since in the past it looked like this stage seemed not to be bitwise reproducible.

After #343 we found out that in fact some reproducibility issues are related just to `torch.save()` but the weights may otherwise be the same.

- [ ] Is this the case for the  Sentence Embdedding training scripts as well (note: test this on just few epochs)?
- [ ] If No, is the training reproducible on CPU at least?
- [ ] If Yes, then track the model trainin with `dvc` by applying the patch.

FrancescoCasalegno:
Blocked, depends on #343

FrancescoCasalegno:
[`torch 1.9.0` just got released](https://pypi.org/project/torch/1.9.0/#files), so this issue is now **unblocked**.

PR #395, opened 2021-07-01
Re-run DVC pipelines to release v0.2.0

pafonta:
Fixes #394.

➡️ READY TO BE MERGED

## Description

- [x] Add the new tag v0.2.0 to `git`.
- [x]  Add the new tag v0.2.0 to each `Dockerfile` for each DVC pipeline.
- [x] Update `whatsnew.rst`, including adding the release date.
- Re-run all DVC pipelines with latest master 12988ef564dd4e6373a7455f5ee30c0608e2e972 (this includes #388):
  - [x] sentence_embedding
  - [x] ner
- Update the `dvc.lock` of all the DVC pipelines:
  - [x] sentence_embedding
  - [x] ner

## Test

This

```
git checkout v0.2.0
dvc pull
dvc repro --recursive .
```

should finish with

```
Data and pipelines are up to date.
```

#### Metrics

For all Transformer-based models and with a precision of `more than 10e-6`, the `pearson_r` has changed. Otherwise, for other models, a lower precision, or other metrics, there is no change.

#### Models

The `sentence_embedding/tf_idf/` model has changed.

As the metrics did not change for this model, we could assume that the binary change for this model comes from the modification in `train.py`:

```
git diff v0.1.1 v0.2.0 -- data_and_models/pipelines/sentence_embedding/train.py
```

#### Metrics

The scores has changed for `DISEASE`, -1.11 % (relative):

| Path                    | Metric           | Old    | New    | Change   |
|-------------------------|------------------|--------|--------|----------|
| disease.json            | entity_f1-score  | 0.7147 | 0.7068 | -0.0079  |
| disease.json            | entity_precision | 0.6976 | 0.6939 | -0.0037  |
| disease.json            | entity_recall    | 0.7327 | 0.7201 | -0.0126  |
| disease.json            | token_f1-score   | 0.7764 | 0.772  | -0.0044  |
| disease.json            | token_precision  | 0.7522 | 0.7504 | -0.0017  |
| disease.json            | token_recall     | 0.8022 | 0.7948 | -0.0074  |
| interrater/disease.json | entity_f1-score  | 0.7346 | 0.7226 | -0.012   |
| interrater/disease.json | entity_precision | 0.6586 | 0.6478 | -0.0108  |
| interrater/disease.json | entity_recall    | 0.8305 | 0.8169 | -0.0136  |

#### Models

All models have changed MD5 sum but not size.

[...]

PR #402, opened 2021-07-14
Fix test_get_grobid_measurements

EmilieDel:
It seems #355 CI started to fail today (14th of July) because of `test_get_grobid_measurements`. Re-triggering the current master is also failing.

## Description

It was impossible for me to reproduce this issue on other platforms (my personal mac laptop and into a docker container). 

From `test_get_grobid_measurements`
- On my personal laptop/docker container
```python
response = requests.Response()
response.status_code = 200
response._content = json.dumps({}).encode("utf-8")

# Results
response.text: '{}'
json.loads(response.text): {}
```
- In the CI:
```python
# Results
response.text: '筽'  # WHY ?
```

From the environment in the CI (yesterday on master - today on this PR), only 3 packages have different versions:

| package | before | after |
| -- | -- | -- |
| platformdirs | 2.0.0 | 2.0.2 |
| python-dateutil | 2.8.1 | 2.8.2 |
| tox | 3.23.1 | 3.24.0  |

Does anyone have an idea of the problem ? 

## How to fix ?

For now, I just replaced the empty dictionary with small dictionary containing one key. Tests are passing now.

### DIFF
--- a/tests/test_mining/test_attribute.py
+++ b/tests/test_mining/test_attribute.py
@@ -807,7 +807,7 @@ class TestAttributeExtraction:
 
         # Test 3
         response.status_code = 200
-        response._content = json.dumps({}).encode("utf-8")
+        response._content = json.dumps({"measurements": []}).encode("utf-8")
         measurements = extractor.get_grobid_measurements(text)
         assert len(measurements) == 0
 


pafonta (approved):
Thank you very much for fixing this!

I have it too yesterday on #395 without knowing why it was suddenly happening.

Note: Today, the tests pass on #395 🤔

Stannislav:
So given the tests pass today, do we still need to do something about this?

EmilieDel:
I don't think so! I will close this PR.

PR #406, opened 2021-07-19
#372 Convert JATS XML to target JSON schema

### LINKED ISSUE #372: Convert JATS XML to target JSON schema
As discussed [here](https://github.com/BlueBrain/Search/issues/298#issuecomment-848197933), JATS XML is the format used by PMC.

PMC seems the most low-hanging fruit to download XML texts, so being able to convert JATS XML to the JSON format specified by #371 is a priority. 

- [ ] Check if CORD-19 made public their tools to convert JATS XML to JSON.
- [ ] If No, then create tools of our own to do so.

EmilieDel:
Fixes #372.

## Description

This PR introduces the class `PubmedXMLParser` to parse `XML` files in the `JATS` format. This covers the open articles that can be downloaded from PubMed Central.

## To Be Discussed

1. We parse the `XML` using `lxml` instead of `xml` [because it is much faster](https://lxml.de/performance.html). However this means that we have yet another dependency in our project, because `lxml` is not in the stdlib.
2. Parsing with `lxml` (or `xml`) is [not approved by `bandit`](https://github.com/SAP/python-pyodata/issues/12) as it may be vulnerable to attacks. Instead, `bandit` recommends using `defusedxml` functions, but since `defusedxml.lxml` is deprecated [it's not clear what is best](https://github.com/SAP/python-pyodata/issues/12#issuecomment-520399969).

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


tests/data/sample_file.xml
```diff
@@ -0,0 +1,193 @@
+<?xml version="1.0"?>
```
jankrepl, line 1: So I guess to create this XML you took a real XLM file from somewhere (?) and modified some of the entries? Or is this an actual template?

requirements.txt
```diff
@@ -23,6 +23,7 @@ en-core-sci-lg @ https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.
 h5py==3.2.1
 ipython==7.22.0
 ipywidgets==7.6.3
+lxml==4.6.3
```
jankrepl, line 26: I guess we want the `requirements.txt` to be sorted alphabetically, right?

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        str
+            The article title.
+        """
+        titles = self.content.find("//title-group/article-title")
+        if titles is None:
+            return ""
+        return self.text_content(titles)
+
+    @property
+    def authors(self) -> Generator[str, None, None]:
```
jankrepl, line 106: So what is the main motivation behind using a `Genarator`? (Also done for `abstract` and `paragraphs`).
On the outside, it might make the user believe that there is "lazy" parsing taking place. However, I quickly checked the methods `xpath` and `findall` and they seem to return lists.


Note that this is just a minor comment and I am totally fine with the way it is

jankrepl (commented):
Looks great! Thank you for the work! I left a couple of comments in the code!

requirements.txt
EmilieDel: Good catch! Done in 580adf9.

tests/data/sample_file.xml
EmilieDel: Yes this is real XML file that was modified. The structure was found on https://jats.nlm.nih.gov/archiving/.

jankrepl: Cool. Thanks!

src/bluesearch/database/article.py
EmilieDel: I would say it was mainly to keep consistency with what was done in `CORD19ArticleParser`.

jankrepl: Yeh, you are right that my remark also applies to the `CORD19ArticleParser`. @Stannislav your take on this?

Stannislav: So in `CORD19ArticleParser` it was just for convenience to avoid caching all items in a list before returning them.

The base class `ArticleParser` has the return type `Iterable[str]`, so it's not at all necessary to use generators, anything that's iterable will do.

jankrepl: I feel like what we are doing is very similar to
```python
l = [1, 2, 3, 4]  
g = (x for x in l) 
```
Seems like false advertizing for lazy evaluation and memory efficiency.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+class PubmedXMLParser(ArticleParser):
+    """Parser for PubMed XML files using the JATS Journal Publishing DTD.
+
+    Parameters
+    ----------
+    path
+        The path to the XML file from PubMed.
+    """
+
+    def __init__(self, path: Union[str, Path]) -> None:
```
jankrepl, line 88: One thing to consider here is that the `CORD19ArticleParser` expects a `dict` in the constractor. Which assumes that the user already loaded the `.json` file. 

Here, we are passing a path. I think we should unify this somehow.

Personally, I would vote for passing paths everywhere. The main reason for it is that we could construct the parsers programmatically.

```python
def construct_parser(parser, path, **kwargs):
      parser_cls = getattr("bluesearch.database.article", parser)
      return parser_cls(path, **kwargs)
```

EmilieDel: Yes we saw that. And @Stannislav mentioned the fact that testing was more complicated if we are passing a path too. But I think he had also other arguments, right @Stannislav ?

jankrepl: @Stannislav hehe, check out this #412, specifically the `add` command. TLDR the goal is to write a CLI that takes
* parser class name
* path to file to be parsed
and then internally parses that file and puts it in the database. However, if the constructor of different `ArticleParser` subclasses differs, then you would need to have a special logic for each of them

jankrepl: IMO parsing should take the raw files and spit out nice Python objects.

The following examples come to mind
```
pd.read_csv(path)
pd.read_pkl(path)
```

Regarding testing, we literally have a bunch of `.json` files inside of `tests/data/...` so in a way it is simpler to just provide their paths than creating a derived fixture that `json.load` them.

I guess what might be annoying is writing a Python `dict` in the source code and then having to `json.dump` it and remember the path. However, I do not think we have tests like this.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+                    caption = ""
+                yield "Table Caption", caption
+
+    @staticmethod
+    def text_content(element: etree._Element) -> str:
+        """Extract all text of an element and of its descendants (at any depth).
+
+        Parameters
+        ----------
+        element
```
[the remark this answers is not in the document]
EmilieDel: Done in c8c8864

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+
+        Returns
+        -------
+        text
```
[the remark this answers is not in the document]
EmilieDel: Sorry I misunderstood the way we need to handle this! Done in c8c8864.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+                # e.g. it could be an organization. We decide to skip those.
+                continue
+
+    @property
+    def abstract(self) -> Generator[str, None, None]:
+        """Get a sequence of paragraphs in the article abstract.
+
+        Yields
+        ------
+        paragraph_text
```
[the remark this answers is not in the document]
EmilieDel: Done in c8c8864

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        """Get all paragraphs and titles of sections they are part of.
+
+        Paragraphs can be parts of text body, or figure or table captions.
+
+        Yields
+        ------
+        section_title
+            The section title.
+        paragraph_text
+            The paragraph content.
```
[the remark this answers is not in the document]
EmilieDel: Done in c8c8864

tests/test_database/test_article.py
```diff
@@ -33,6 +41,58 @@ def paragraphs(self):
         yield from self._paragraphs
 
 
+@pytest.fixture(scope="session")
+def pubmed_xml_parser():
+    path = pathlib.Path(__file__).parent.parent / "data" / "sample_file.xml"
+    parser = PubmedXMLParser(path.absolute())
```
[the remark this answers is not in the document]
EmilieDel: Done in 8cd70d6

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        tables = self.content.xpath("//body//table-wrap")
+        if tables is not None:
+            for table in tables:
+                if table.find("caption/p") is not None:
+                    caption = self.text_content(table.find("caption/p"))
+                elif table.find("caption/title") is not None:
+                    caption = self.text_content(table.find("caption/title"))
+                else:
+                    caption = ""
+                yield "Table Caption", caption
```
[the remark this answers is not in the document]
EmilieDel: Done in be3cbf0

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        for paragraph in paragraphs:
+            text = self.text_content(paragraph)
+            section = paragraph.find("../title")
+            if section is not None:
+                section = self.text_content(section)
+            yield section or "", text
+
+        # Figure captions
+        figs = self.content.findall("//body//fig")
+        if figs is not None:
```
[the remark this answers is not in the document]
EmilieDel: Done in be3cbf0

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        Parameters
+        ----------
+        element
+
+        Returns
+        -------
+        text
+            Entire text of the element and its descendants.
+        """
+        return "".join(t for t in element.itertext())
```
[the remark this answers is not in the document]
EmilieDel: Done in be3cbf0

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        """Get a sequence of paragraphs in the article abstract.
+
+        Yields
+        ------
+        paragraph_text
+            The paragraphs of the article abstract.
+        """
+        abstract_pars = self.content.findall("//abstract//p")
+        for paragraph in abstract_pars:
+            yield self.text_content(paragraph) or ""
```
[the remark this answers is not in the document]
EmilieDel: Done in be3cbf0

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+            The paragraph content.
+        """
+        # Paragraphs of text body
+        paragraphs = self.content.xpath("//body//p[not(parent::caption)]")
+        for paragraph in paragraphs:
+            text = self.text_content(paragraph)
+            section = paragraph.find("../title")
+            if section is not None:
+                section = self.text_content(section)
+            yield section or "", text
```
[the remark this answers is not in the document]
EmilieDel: Done in be3cbf0

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        # Figure captions
+        figs = self.content.findall("//body//fig")
+        if figs is not None:
+            for fig in figs:
+                try:
+                    fig_captions = fig.find("caption").getchildren()
+                    caption = " ".join([self.text_content(c) for c in fig_captions])
+                    yield "Figure Caption", caption
+                except AttributeError:
+                    continue
```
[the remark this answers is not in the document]
EmilieDel: Done in dffb4a6

tests/test_database/test_article.py
```diff
@@ -33,6 +41,58 @@ def paragraphs(self):
         yield from self._paragraphs
 
 
+@pytest.fixture(scope="session")
+def pubmed_xml_parser():
+    path = pathlib.Path(__file__).parent.parent / "data" / "sample_file.xml"
```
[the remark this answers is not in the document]
EmilieDel: Done in ab6833c.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,127 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+    """Parser for PubMed XML files using the JATS Journal Publishing DTD.
+
+    Parameters
+    ----------
+    path
+        The path to the XML file from PubMed.
+    """
+
+    def __init__(self, path: Union[str, Path]) -> None:
+        self.content = etree.parse(str(path))  # nosec
```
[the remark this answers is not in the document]
EmilieDel: `lxml.etree.iterparse` is still raising a `bandit` issue:
```python
 Issue: [B410:blacklist] Using iterparse to parse untrusted XML data is known to be vulnerable to XML attacks. Replace iterparse with the equivalent defusedxml package.
   Severity: Low   Confidence: High
   Location: src/bluesearch/database/article.py:24
   More Info: https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html#b410-import-lxml
23      # from lxml import etree  # nosec
24      from lxml.etree import iterparse
```

tests/data/sample_file.xml
```diff
@@ -0,0 +1,193 @@
+<?xml version="1.0"?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.2 20190208//EN" "JATS-journalpublishing1.dtd">
```
pafonta, line 2: Is there a reason to use v1.2 while v1.3 is the current version?

src/bluesearch/database/article.py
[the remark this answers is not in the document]
Stannislav: Wow, crayz. According to [this comment](https://github.com/PyCQA/bandit/issues/435#issuecomment-537865438) ther `resolve_entities=False` parameter should make it safe, but `bandit` still flags it. To be honest, I'm not sure I know a good solution. I guess `# nosec` will do for now.

There's some more details on XML parsing and available python libraries [in these EuroPython slides](https://ep2021.europython.eu/media/conference/slides/BjExfG9-python-security-best-practises.pdf) if anyone wants to dig deeper.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,121 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
```
pafonta, line 89: To actually use inheritance, `super().__init__(...)` should be called.

This is also a best practice to avoid the current case. Indeed, the child classes have a different API. This is a contradiction with the goal of using inheritance.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,121 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        for author in authors:
+            try:
+                given_names = self.text_content(author.find("name/given-names"))
+                surname = self.text_content(author.find("name/surname"))
+                author_str = given_names + " " + surname
+                yield author_str.strip()
+            except AttributeError:
+                # In rare cases, an author may not have a given name or a surname,
+                # e.g. it could be an organization. We decide to skip those.
+                continue
```
Stannislav, line 122: Just wondering - is it the `author.find(...)` that can throw the `AttributeError`? Seems kind of strange.

pafonta (commented):
Thank you for diving into the the JATS DTD.

I have left two comments.

tests/data/sample_file.xml
EmilieDel: This file was taken directly on JATS website. The only thing changed were mainly the metadata inside because it was an real article. I am not sure this version has an impact on the parsing itself.

src/bluesearch/database/article.py
```diff
@@ -73,6 +76,121 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+
+        # Table captions
+        tables = self.content.xpath("//body//table-wrap") or []
+        for table in tables:
+            caption_element = table.find("caption/p") or table.find("caption/title")
+            caption = self.text_content(caption_element)
+            yield "Table Caption", caption
+
+    @staticmethod
+    def text_content(element: etree._Element) -> str:
```
Stannislav, line 175: I think one might need to change the type to `Optional[etree._Element]`

EmilieDel: Done in f8b21a0. Thanks for catching this!

src/bluesearch/database/article.py
EmilieDel: Done in f8b21a0. I forgot to update that part of the code after refactoring `text_content()`

src/bluesearch/database/article.py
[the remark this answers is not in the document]
EmilieDel: Hey @Stannislav, I tried to implement an alternative with `defusedxml` in e9eb447. Do not hesitate to give your feedback!

src/bluesearch/database/article.py
EmilieDel: Done in 098c744

tests/data/sample_file.xml
pafonta: Following our discussion, I have checked what is the JATS version used in PMC. I have used the article put as example for their OA service [here](https://www.ncbi.nlm.nih.gov/pmc/utils/oa/oa.fcgi?id=PMC5334499). The version is v1.0, which is lower than the one currently used in the parser of the PR. There might then be some issues.

```xml
<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.0 20120330//EN" "JATS-archivearticle1.dtd">
```

EmilieDel: Good point, @pafonta ! I just checked with the data we have for now. It seems that several versions are used depending on the article. Check some stats I found here:

| Version | Number of Articles |
| ---- | ---- |
| Journal Archiving and Interchange DTD v1.0 | 826'545 |
| Journal Archiving and Interchange DTD v2.3 | 73'875 |
| Journal Archiving and Interchange DTD v3.0 | 1 |

EmilieDel: After checking manually, it seems we are able to extract all informations needed (`title`, `authors`, `abstract` and `paragraphs` - also `figure` and `table`) no matter the version of the schema.

src/bluesearch/database/article.py
pafonta: The complete implementation will be part of https://github.com/BlueBrain/Search/issues/419.

Indeed:
- the API of parser classes will be harmonized,
- there will be a constructor in the parent class,
- it will be called through `super()`.

tests/data/sample_file.xml
pafonta: Thank you @EmilieDel for checking.

JATS v2.3 and 3.0 do not exist, so I guess they are incorrect data.

src/bluesearch/database/article.py
Stannislav: So I guess #419 is the followup to this?

src/bluesearch/database/article.py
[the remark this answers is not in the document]
Stannislav: awesome, thanks!

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        for table in tables:
+            caption_elements = table.findall(".//caption/p") or table.findall(
+                ".//caption/title"
+            )
+            if caption_elements is None:
+                continue
+            caption = " ".join(self.text_content(c) for c in caption_elements)
+            yield "Table Caption", caption
+
+    def get_sections(self) -> dict:
```
Stannislav, line 199: Maybe it's possible to add a more precise annotation like `Dict[???, str]` here?

Seems like the keys are some kind of objects? It looks like the type is something that `self.content.findall` returns an iterable of, but I wasn't able to determine the type by just reading the code.

src/bluesearch/database/article.py
EmilieDel: Yes #419 is the followup.

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        """Extract sections information.
```
Stannislav, line 200: So, after reading the implementation, this function returns the mapping
```
paragraph_object -> section title
```
Is this true?

So basically go through the whole article and for every paragraph find the corresponding section title?

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        sections_dir : dict
+            Dictionary whose keys are paragraphs and value are section title.
+        """
+        sections_dir = {}
+        for sec in self.content.findall(".//sec"):
+
+            section_title = ""
+            section_paragraphs = []
+
+            for element in list(sec):
```
Stannislav, line 213: It seems `sec` is already an iterable, so maybe casting to `list` is not necessary?

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+
+                if element.tag == "title":
+                    section_title = self.text_content(element)
+                elif element.tag == "p":
+                    section_paragraphs.append(element)
+                else:
+                    continue
```
Stannislav, line 220: I think these two lines can be dropped.

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+            The section title.
+        text : str
+            The paragraph content.
+        """
+        # Paragraphs of text body
+        abstract_p = self.content.findall(".//abstract/p")
+        caption_p = self.content.findall(".//caption/p")
+        acknowlegd_p = self.content.findall(".//ack/p")
+
+        exclude_list = abstract_p + caption_p + acknowlegd_p
```
Stannislav, line 168: What are those `_p` paragraphs and why do we want to exclude them?

Stannislav (commented):
Just left some more comments.

Also I think it's probably reallly hard to test if the parser works correctly for all edge cases. Maybe we can try and run it on real data and manually check if the results make sense.

src/bluesearch/database/article.py
EmilieDel: - Abstract is found in the `abstract` property of the class. So it is to avoid duplicates.
- Caption is saved after and keep the section name `Figure Caption` or `Table Caption`. So it is to avoid duplicates.
- Acknowledgements are the acknowledgements of the paper, I don't think we need to keep those paragraphs.

src/bluesearch/database/article.py
EmilieDel: Yes! I didn't find another solution because `xml` parsing only allowed to find children of an object, not the parents!

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+        ------
+        section : str
+            The section title.
+        text : str
+            The paragraph content.
+        """
+        # Paragraphs of text body
+        abstract_p = self.content.findall(".//abstract/p")
+        caption_p = self.content.findall(".//caption/p")
+        acknowlegd_p = self.content.findall(".//ack/p")
```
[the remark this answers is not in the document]
EmilieDel: It is for `acknowledgements` but I didn't want to keep the entire word. I find this enough explicite for everyone to understand what it is.

src/bluesearch/database/article.py
```diff
@@ -73,6 +87,163 @@ def paragraphs(self) -> Iterable[Tuple[str, str]]:
...
+                continue
+            caption = " ".join(self.text_content(c) for c in caption_elements)
+            yield "Table Caption", caption
+
+    def get_sections(self) -> dict:
+        """Extract sections information.
+
+        Returns
+        -------
+        sections_dir : dict
```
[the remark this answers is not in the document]
EmilieDel: I removed it.

src/bluesearch/database/article.py
EmilieDel: Done in d0db1e4.

src/bluesearch/database/article.py
EmilieDel: Done in d0db1e4

src/bluesearch/database/article.py
EmilieDel: Done in d0db1e4.

src/bluesearch/database/article.py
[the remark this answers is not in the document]
EmilieDel: Done in cf92952.

src/bluesearch/database/article.py
Stannislav: Ah I see :) Would it be possible to leave some comments in the code along the same lines? I think this could help to read the code later.

src/bluesearch/database/article.py
Stannislav: I see, thanks! Could we maybe adjust the method title and the docstring to make this more clear? I think `get_sections` might be slightly misleading because we don't really return sections. Maybe also the docstring (title?) could mention it actually constructs a mapping `paragraph element -> section title`.

src/bluesearch/database/article.py
EmilieDel: Done in a4f7a73

src/bluesearch/database/article.py
EmilieDel: Done in a4f7a73

Stannislav:
I just ran the parser on the XML file in the tests (`sample_file.xml`) by hand, I'm getting these paragraphs:
```
[('', 'Paragraph 1'),
 ('', 'Paragraph 2'),
 ('', 'Paragraph 3'),
 ('Section Title 1', 'Paragraph Section 1'),
 ('Section Title 2', 'Paragraph Section 2'),
 ('Section Title 2', 'Paragraph 2 Section 2'),
 ('Discussion', 'Paragraph Discussion'),
 ('Figure Caption', '\n                    Figure 1 Caption\n                '),
 ('Table Caption', 'Table 1 Caption'),
 ('Table Caption', 'Table 2 Caption'),
 ('Table Caption', '')]
```
so, I'm wondering:
- Is it a problem that the first three paragraphs don't have section titles? Is something in our pipeline going to break because of this?
- Do we know where the newlines / spaces in the figure caption come from? I looked in the XML and I can't seem them.
- The last paragraph is empty, is this  a problem?

EmilieDel:
Hello @Stannislav, 

Thanks for checking this! 

First, here are my results:
```
[('', 'Paragraph 1'),
 ('', 'Paragraph 2'),
 ('', 'Paragraph 3'),
 ('Section Title 1', 'Paragraph Section 1'),
 ('Section Title 2', 'Paragraph Section 2'),
 ('Section Title 2', 'Paragraph 2 Section 2'),
 ('Discussion', 'Paragraph Discussion'),
 ('Figure Caption', '\nFigure 1 Caption\n'),
 ('Table Caption', 'Table 1 Caption'),
 ('Table Caption', 'Table 2 Caption'),
 ('Table Caption', 'Table 3 Caption')]
```

1. Paragraphs are not always under a section. Normally the pipeline should not break because of this. 
2. This one is strange, I will investigate to see where it comes from.
3. I don't obtain the same results as you for that one.

Stannislav:
@EmilieDel wow, weird. I just checked - I'm on the latest commit (a4f7a73). Here's the entire set of commands I ran:
```python
$ python
Python 3.9.5 (default, Jun  9 2021, 14:27:46) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pprint import pprint
>>> from bluesearch.database.article import PubmedXMLParser, Article
>>> parser = PubmedXMLParser("tests/data/sample_file.xml")
>>> parser.title
'Article Title'
>>> list(parser.authors)
['Author Given Names 1 Author Surname 1', 'Author Given Names 2 Author Surname 2']
>>> list(parser.abstract)
['Abstract Paragraph 1', 'Abstract Paragraph 2']
>>> pprint(list(parser.paragraphs))
[('', 'Paragraph 1'),
 ('', 'Paragraph 2'),
 ('', 'Paragraph 3'),
 ('Section Title 1', 'Paragraph Section 1'),
 ('Section Title 2', 'Paragraph Section 2'),
 ('Section Title 2', 'Paragraph 2 Section 2'),
 ('Discussion', 'Paragraph Discussion'),
 ('Figure Caption', '\n                    Figure 1 Caption\n                '),
 ('Table Caption', 'Table 1 Caption'),
 ('Table Caption', 'Table 2 Caption'),
 ('Table Caption', '')]
>>>
```
just ran it again and am still getting the same output.

EmilieDel:
@Stannislav I am able to reproduce your issue! I am checking it now.

EmilieDel:
@Stannislav I found the problem, I am testing against a slightly different `sample_file.xml` locally. There is no caption for `Table 3` in the `tests/data/sample_file.xml` we have now on the last commit. See here: 
```xml
 <table-wrap id="T3">
                    <label>Table 3</label>
                    <caption>
                    </caption>
                    <table>
                    </table>
                </table-wrap>
```

So the behaviour sounds good.

Stannislav (approved):
Thanks a lot.

I'm thinking about these potential follow-ups:
- Disallow empty paragraphs? (like in the discussion where a table caption was empty)
- Dealing with significant spaces. `strip()` might already do a good job, but I'm wondering what would happen to content like this:
  ```xml
  <abstract>
      <p id="P1">While previous studies have shown that CD4+ T cells expressing CCR6 and
          CD161 are depleted from blood during HIV infection, the mechanisms underlying
          their loss remain unclear. In this study, we investigated how the homeostasis of
          CCR6+ and CD161+ CD4+ T cells contributes to SIV disease progression and the
          mechanisms responsible for their loss from circulation. By comparing SIV
          infection in rhesus macaques (RMs) and natural host sooty mangabeys (SMs), we ...
      </p>
  </abstract>
  ```
  `strip()` won't remove the intermediate indentations, right? Maybe this is something that never happens with PMC files though, I don't know.
- We use `element.find(".//some/path")`, [the double-slash is a glob for all elements at all sub-levels](https://docs.python.org/3/library/xml.etree.elementtree.html#xpath-support). If we know the exact (fixed) structure of documents, we might consider using a single slash instead. Might be worth checking the performance impact.
- LaTeX code / formulas. See `PMC7122938` for an example. The formula `R_0 >= 1` leads to this block in the XML files:
  ```xml
  <inline-formula id="IEq3">
      <alternatives>
          <tex-math id="M5">\documentclass[12pt]{minimal}
              \usepackage{amsmath}
              \usepackage{wasysym}
              \usepackage{amsfonts}
              \usepackage{amssymb}
              \usepackage{amsbsy}
              \usepackage{mathrsfs}
              \usepackage{upgreek}
              \setlength{\oddsidemargin}{-69pt}
              \begin{document}$$ R_{0} \ge 1 $$\end{document}
          </tex-math>
          <mml:math id="M6">
              <mml:mrow>
                  <mml:msub>
                      <mml:mi>R</mml:mi>
                      <mml:mn>0</mml:mn>
                  </mml:msub>
                  <mml:mo>&#x02265;</mml:mo>
                  <mml:mn>1</mml:mn>
              </mml:mrow>
          </mml:math>
          <inline-graphic xlink:href="466534_1_En_109_Chapter_IEq3.gif"/>
      </alternatives>
  </inline-formula>
  ```
  so what to do with this? Maybe just throw away? Maybe replace it with some generic placeholder token like usually done in ML?
- Tags of the form `prefix:sometag` (can be seen in the example above) are called namespaces. [There's some support in the `xml` library for handling them](https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml-with-namespaces). Maybe it could be useful, maybe not.

EmilieDel:
Hello @Stannislav, 

FYI, I created #425, #427, #428 & #429 followup issues to your comment. Thanks for all the suggestions of improvements!

jankrepl (approved):
LGTM! Thank you for all the work:)

PR #417, opened 2021-08-04
Make entrypoint startup faster

### LINKED ISSUE #411: Entrypoint startup overhead
## 🐛 Bug description
Currently, all our entrypoints are lying inside of the folder `src/bluesearch/entrypoint/`. This folder also contains `__init__.py` that actually imports all the functions from `src/bluesearch/entrypoint/*.py` modules.

**PRO**
This makes the path to the function shorter `from bluesearch.entrypoint import foo` instead of `from bluesearch.entrypoint.bar import foo`.

**CON**
Launching of a single entrypoint will trigger the import of **all** entrypoints. This is not great because some of our entrypoints are slow to import.


I would argue that the above mentioned **pro** is irrelevant for entrypoints. We only hardcode the full path once inside
of `setup.py` when defining `console_scripts` and potentially inside some tests. The entrypoints are not supposed to be imported in a Python session.

## To reproduce
This should work if you have an editable install of `bluesearch`
```bash
LOC="src/bluesearch/entrypoint/"                                                                                                                                                
SCRIPT="bluesearch.entrypoint.compute_embeddings"  # feel free to change to a different entrypoint                                                                                                                             
                                                                                                                                                                                
time python -m $SCRIPT -h > /dev/null
                                                                                                                                           
mv $LOC'__init__.py' $LOC'__init__backup.py'                                                                                                                                    
touch $LOC'__init__.py'                                                                                                                                                         
                                                                                                                                                                                
time python -m $SCRIPT -h > /dev/null
                                                                                                                                           
mv $LOC'__init__backup.py' $LOC'__init__.py'
```
```bash
real    0m3.441s
user    0m2.189s
sys     0m0.578s

real    0m0.349s
user    0m0.314s
sys     0m0.073s
```

## Suggested solution
Erase everything inside of `src/bluesearch/entrypoint/__init__.py` and always refer to the entrypoints with the full path.

Stannislav:
Fixes #411.

## Description
Our CLI entrypoints have a very slow startup time. This is because all entrypoint submodules are imported in `bluesearch.entrypoint.__init__`.

The problem is solved by removing these imports.

## How to test?

- `pytest tests/test_entrypoint`
- Try running the entrypoint in the shell, observe much faster startup time.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


EmilieDel (approved):
LGTM!

jankrepl (approved):
Great! Tested it and now each entrypoint has a different startup time which makes perfect sense:)

Issue #419, opened 2021-08-11
Harmonize parser's constructor


EmilieDel:
## 🚀 Feature

`CORD19ArticleParser` and `PubmedXMLParser` are classes inheriting from the abstract class `ArticleParser`. It could be nice:
- Harmonise the constructors
- Create a constructor in the `ArticleParser` class

## Motivation

It is going to be easier/cleaner for the entrypoint parsing data.

Stannislav:
Discussed in the last planning: currently the different constructors don't pose any problems. Come back to this issue if it starts being relevant again.

PR #434, opened 2021-09-01
Add support for parsing PubMed XMLs from command line

### LINKED ISSUE #433: Add support for `PubmedXMLParser` in `bbs_database parse` CLI
## Context
 In #412 the first version of the `bbs_database parse` command was created. It was decided (see discussion [here](https://github.com/BlueBrain/Search/pull/412#discussion_r681066565)) to avoid dynamic resolution of the `--parser` argument, therefore relying on a sequence of `if ... else` statements.

## Issue
 In #406 we added support for the `PubmedXMLParser` parser, but `bbs_database parse` still does not support it. So we need to add support for it.

FrancescoCasalegno:
Fixes #433.

## How to test?
First, try the following snippet to see the available parameters for the parser.
```bash
$ bbs_database parse -h
usage: bbs_database parse [-h]
                          {CORD19ArticleParser,PubmedXMLParser} input_path
                          output_path

Parse article.
...
```
Second, try to parse a PMC XML file.
```bash
$ bbs_database parse PubmedXMLParser tests/data/sample_file.xml sample_file.pkl
```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
- [x] All CI tests pass.


src/bluesearch/entrypoint/database/parse.py
```diff
@@ -40,9 +46,12 @@ def run(
     Parameter description and potential defaults are documented inside of the
     `get_parser` function.
     """
+    parser_inst: ArticleParser
```
jankrepl, line 49: Shouldn't `mypy` infer this automatically?

jankrepl (approved):
Looks great! Thank you!

tests/entrypoint/database/test_parse.py
```diff
@@ -1,15 +1,21 @@
...
-        main(["parse", "WrongParser", "path_to_input", "path_to_output"])
+    wrong_parser = "WrongParser"
+    try:
+        main(["parse", wrong_parser, "path_to_input", "path_to_output"])
+
+    # argparse exists with error 2, so we need to "unpack" the exception
+    except SystemExit as e:
+        context = e.__context__
+        assert isinstance(context, ArgumentError)
+        assert f"invalid choice: '{wrong_parser}'" in context.args[1]
```
Stannislav, line 18: Isn't this going to pass if no `SystemExit` is raised?

Maybe we can do something like this:
```suggestion
    wrong_parser = "WrongParser"
    with pytest.raises(SystemExit) as exc_info:
        main(["parse", wrong_parser, "path_to_input", "path_to_output"])

    # argparse exits with error code 2, so we need to "unpack" the exception
    exc = exc_info.value
    context = exc.__context__
    assert exc.code == 2
    assert isinstance(context, ArgumentError)
    assert f"invalid choice: '{wrong_parser}'" in context.message
```

Stannislav (commented):
Thanks :)

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: I totally agree that it should, but unfortunately it doesn't. Without it, `mypy` thinks that the same variable assigned tow lines above may be re-assigned now to a different type, even if there's a clear `elif` in the middle. 

So I have to leave this type definition here or `mypy` will rant as follows:
```python
src/bluesearch/entrypoint/database/parse.py:48: 
error: Incompatible types in assignment (expression has type "PubmedXMLParser", variable has type "CORD19ArticleParser")  [assignment]
```

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -14,6 +19,7 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "parser",
         type=str,
+        choices=("CORD19ArticleParser", "PubmedXMLParser"),
```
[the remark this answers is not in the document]
FrancescoCasalegno: I agree with you that it's a good idea, the only thing that stopped me before was that this would add a further layer of unnecessary (?) abstraction.
But yes, from the user perspective `cord19-json` is probably much easier to remember than `CORD19ArticleParser`.

Fixed in 82700402c359a1593904556aa286086fb52a2d9d.

tests/entrypoint/database/test_parse.py
FrancescoCasalegno: Thanks for the suggestion! Fixed in 8270040.

The only thing that I left as it was is the `context.args[1]` instead of `context.message`. The `message` attribute of `BaseException` was deprecated in Python 2.6 and removed in Python 3.  Now, it is true that some exception types like `ArgumentError` have their own` .message`, but `mypy` and PyCharm don't know what's the type of `context` so they complain about it.

Stannislav: Interesting, they do? I would have assumed that `mypy` and `PyCharm` would infer from `assert isinstance(context, ArgumentError)` that `context` is an instance of `ArgumentError`...

Stannislav: Actually, I just tried, `mypy` seems to be fine with that. PyCharm complains with the message `Unresolved attribute reference 'message' for class 'ArgumentError'`. Looks like a bug to me.

Another option is to write
```python
assert f"invalid choice: '{wrong_type}'" in str(context)
```

Anyway, not so important. We can resolve :)

Stannislav (approved):

pafonta (commented):
That works.

During a planning, we decided to tackle this issue with https://github.com/BlueBrain/Search/issues/419. Has something changed since then?

tests/entrypoint/database/test_parse.py
FrancescoCasalegno: > I would have assumed that `mypy` and PyCharm would infer from `assert isinstance(context, ArgumentError)` that context is an instance of `ArgumentError`

You're right! I tried w/o the `assert isinstance` and then `mypy` complains, but with that `assert` everything is fine.

Just in case, also to make PyCharm happy, I switched to the `str(context)` you suggested in your last message.

Fixed in 880176ccea2516298652915a524a470dde7cb920.

PR #440, opened 2021-09-13
Tables articles and sentences for bbs_database init and add

### LINKED ISSUE #435: Populate more columns and tables with `bbs_database add`
### Context
After #412, we have added support for two CLI operations.
  - `bbs_database parse` – calls a parser (an instance of a class which should inherit from [`ArticleParser`](https://github.com/BlueBrain/Search/blob/86ba479406020605aac5de19fdfb93437a014a15/src/bluesearch/database/article.py#L42)) on the file name, and dumps the whole parsed article into a pickled file.
  - `bbs_database add` – loads the pickled file and uses that `Article` instance to `INSERT` values into tables of a database.

Currently the only table we create is `articles`, and the only column that we fill is `title`:
https://github.com/BlueBrain/Search/blob/a37bfca303728e3a8cfe94abff533584b44e04ea/src/bluesearch/entrypoint/database/init.py#L58-L65

But, as we did for the CORD19 database, we need another table for the sentences and way more columns:
https://github.com/BlueBrain/Search/blob/138f689f06ba1a8216f62579327a12afd4652ded/src/bluesearch/database/cord_19.py#L146-L204

### Actions
Create a PR that
- [x] Proposes a schema for the tables `articles` and `sentences` created with `bbs_database init` — it doesn't need to be the definitive one, but at least start adding the columns that we know are for sure relevant.
- [x] Populates the fields of table with `bbs_database add`.

## Motivation
Once we have more fields populated, we will also be able to identify more issues and create new tickets accordingly — e.g. broken parsing, (almost-)empty sentences, too long sentences, ...

pafonta:
Fixes #435.

## Description

Add the tables `articles` and `sentences` for the entrypoint `bbs_database` and its subcommands `init` and `add`.

## How to test?

```
bbs_database init test.db
bbs_database parse pmc-xml tests/data/sample_file.xml article.pkl
bbs_database add test.db article.pkl
```

## InvocationError

Since the merging of #438 in master and then in this PR, the test suite is failing.

The reason is the following `InvocationError` from `tox`:
```
tests/integration/test_bbs_database.py::test_bbs_database[mysql] ERROR: InvocationError for command /home/runner/work/Search/Search/.tox/py37/bin/pytest -m '' --color=yes (exited with code -9 (SIGKILL)) (exited with code -9)
```

A strange thing is that it happens only on `Python 3.7`, `Ubuntu 20.04`, and `MySQL` in the GitHub CI. Besides, investigating a `SIGKILL` in Python code could be tricky. Indeed, this would imply an error in some low-level C or C++ code.

Several debugging attempts were made to identify the cause of this `InvocationError`. A number of solutions were implemented (see collapsed part below). The corresponding HEAD is ded804c.

However, the next day, modifying a comment in 0e0a84e made the test suite fail again. The only difference between the two test suites is the version of the GitHub runner used. The conclusion [here](https://github.com/BlueBrain/Search/pull/440#issuecomment-925832410) was finally drawn. It was decided then to skip the `test_bbs_database` on `Ubuntu` and `Python 3.7` and to revert all the changes below.

➡️  A follow-up issue was created: #456.

<details>
  <summary>The implemented solutions were:</summary>

To sum up, there were 3 causes:
1. One should  connect to `MySQL` with `charset=utf8mb4` (even with the driver `pymysql` and not `mysqldb`).
2. One shouldn't use `Language.pipe()` from `spaCy` to speed-up NLP operations.
3. One should use `str(x)` instead of `x.text` with `x` being an element of `Doc.sents` from `spaCy`.

Implementing 1 and 3 do not have an immediate impact. However, implementing 2 does. Indeed, the NLP operations would be slower. With the current code structure, there is no solution without a refactoring. That's why an issue dedicated to using `Language.pipe()` to split paragraphs into sentences has been created: #454.

The fixes could be consulted with:
```
git diff 0c39084 ded804
```
and gives:
```
git diff 0c39084 -- src/bluesearch/entrypoint/database/add.py
diff --git a/src/bluesearch/entrypoint/database/add.py b/src/bluesearch/entrypoint/database/add.py
index dca28e2..a1b88ca 100644
--- a/src/bluesearch/entrypoint/database/add.py
+++ b/src/bluesearch/entrypoint/database/add.py
@@ -75,7 +75,8 @@ def run(
         engine = sqlalchemy.create_engine(f"sqlite:///{db_url}")
 
     elif db_type == "mysql":
-        engine = sqlalchemy.create_engine(f"mysql+pymysql://{db_url}")
+        # Not using 'charset=utf8mb4' gives an issue on Python 3.7 and Ubuntu 20.04.
+        engine = sqlalchemy.create_engine(f"mysql+pymysql://{db_url}?charset=utf8mb4")
 
     else:
         # This branch never reached because of `choices` in `argparse`
@@ -114,15 +115,14 @@ def run(
         }
         article_mappings.append(article_mapping)
 
-        swapped = (
-            (text, (section, ppos))
-            for ppos, (section, text) in enumerate(article.section_paragraphs)
-        )
-        for doc, (section, ppos) in nlp.pipe(swapped, as_tuples=True):
+        # Using spaCy Language.pipe() gives an issue on Python 3.7 and Ubuntu 20.04.
+        for ppos, (section, text) in enumerate(article.section_paragraphs):
+            doc = nlp(text)
             for spos, sent in enumerate(doc.sents):
                 sentence_mapping = {
                     "section_name": section,
-                    "text": sent.text,
+                    # Using 'sent.text' gives an issue on Python 3.7 and Ubuntu 20.04.
+                    "text": str(sent),
                     "article_id": article_id,
                     "paragraph_pos_in_article": ppos,
                     "sentence_pos_in_paragraph": spos,
```

</details>

## Checklist

- [x] All CI tests pass.


src/bluesearch/entrypoint/database/schemas.py
```diff
@@ -0,0 +1,73 @@
...
+    UniqueConstraint,
+)
+
+
+def schema_articles(metadata: MetaData) -> None:
+    """Add to the given 'metadata' the schema of the table 'articles'."""
+    Table(
+        "articles",
+        metadata,
+        Column("article_id", String(32), primary_key=True),
```
jankrepl, line 39: Why 32?

jankrepl (commented):
Great:)

Maybe you could add some tests/modify existing tests inside of `test_init.py` and `test_add.py`?

Some ideas:
* (`test_init.py`) Make sure that the 2 tables `articles` and `sentences` are there with the correct columns (after initialization)
* (`test_add.py`) Make sure that the `sentences` table is nonempty after (after adding the articles)

Stannislav:
I just ran the CLI commands from the description and I'm getting warnings + errors (python3.9):
```
$ bbs_database add test.db article.pkl
[2021-09-15 11:12:03,537] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,539] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '1'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,539] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,539] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '2'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,540] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,540] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '3'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,540] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Section'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '1'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Section'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '2'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '2'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Section'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '2'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Paragraph'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,541] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Discussion'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Figure'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '1'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Caption'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Table'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '1'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Caption'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Table'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token '2'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
[2021-09-15 11:12:03,542] [WARNING] [W108] The rule-based lemmatizer did not find POS annotation for the token 'Caption'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1771, in _execute_context
    self.dialect.do_execute(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 717, in do_execute
    cursor.execute(statement, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: sentences.article_id, sentences.paragraph_pos_in_article, sentences.sentence_pos_in_paragraph

The above exception was the direct cause of the following exception:

  File "/Users/sschmidt/Code/BlueBrain/Search/venv/bin/bbs_database", line 33, in <module>
    sys.exit(load_entry_point('bluesearch', 'console_scripts', 'bbs_database')())
  File "/Users/sschmidt/Code/BlueBrain/Search/src/bluesearch/entrypoint/database/parent.py", line 63, in main
    command_map[command](**kwargs)  # type: ignore
  File "/Users/sschmidt/Code/BlueBrain/Search/src/bluesearch/entrypoint/database/add.py", line 114, in run
    con.execute(query, sentence_mapping)
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1263, in execute
    return meth(self, multiparams, params, _EMPTY_EXECUTION_OPTS)
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/sql/elements.py", line 323, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1452, in _execute_clauseelement
    ret = self._execute_context(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1814, in _execute_context
    self._handle_dbapi_exception(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1995, in _handle_dbapi_exception
    util.raise_(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/util/compat.py", line 207, in raise_
    raise exception
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 1771, in _execute_context
    self.dialect.do_execute(
  File "/Users/sschmidt/Code/BlueBrain/Search/venv/lib/python3.9/site-packages/sqlalchemy/engine/default.py", line 717, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: sentences.article_id, sentences.paragraph_pos_in_article, sentences.sentence_pos_in_paragraph
[SQL: INSERT INTO sentences(section_name, text, article_id, paragraph_pos_in_article, sentence_pos_in_paragraph) VALUES(?, ?, ?, ?, ?)]
[parameters: ('Table Caption', 'Table 2 Caption', 'd9ec6b52136a2a3f396ef4a7c43a94f4', 9, 0)]
(Background on this error at: https://sqlalche.me/e/14/gkpj)
```
Has anyone has similar issues?

src/bluesearch/entrypoint/database/add.py
```diff
@@ -62,6 +65,50 @@ def run(
...
+    for pposition, (document, section) in enumerate(nlp.pipe(swapped, as_tuples=True)):
+        for sposition, sentence in enumerate(document.sents):
+            sentence_mapping = {
+                "section_name": section,
+                "text": sentence.text,
+                "article_id": article_id,
+                "paragraph_pos_in_article": pposition,
+                "sentence_pos_in_paragraph": sposition,
+            }
+        sentence_mappings.append(sentence_mapping)
```
Stannislav, line 103: Maybe this was meant to be insde the inner for-loop?

src/bluesearch/entrypoint/database/add.py
```diff
@@ -62,6 +65,50 @@ def run(
...
+
+    sentences_keys = sentence_mapping.keys()
```
Stannislav, line 105: `sentence_mapping` might not be defined at this point (if the for-loop is zero length)

src/bluesearch/entrypoint/database/add.py
```diff
@@ -62,6 +65,50 @@ def run(
...
+        "article_id": article_id,
+        "title": article.title,
+        "authors": ", ".join(article.authors),
+        "abstract": "\n".join(article.abstract),
+    }
+    article_keys = article_mapping.keys()
+    article_fields = ", ".join(article_keys)
+    article_binds = f":{', :'.join(article_keys)}"
+
+    with engine.connect() as con:
```
Stannislav, line 83: Maybe we can consider using `engine.begin()` to make use of [transactions](https://docs.sqlalchemy.org/en/14/core/connections.html#using-transactions)? If I read the SQLAlchemy documentation correctly it's safer for when an exception is raised within the `with`-block.

Stannislav: There's also another `engine.connect()` further below, I guess the same applies there too.

src/bluesearch/entrypoint/database/add.py
```diff
@@ -4,6 +4,9 @@
 
 import sqlalchemy
 
+from bluesearch.database.identifiers import generate_uid
+from bluesearch.utils import load_spacy_model
+
```
Stannislav, line 9: Maybe we should discuss our best practises here, I think we can consider running all import within the actual entry-point function, i.e. inside the `run`-function.

I'm mentioning this because inside `parent.py` we construct the CLI by importing all subparsers and so all imports in the module scope of the subparser module will be run, which will eventually slow down the whole CLI initialization.

src/bluesearch/entrypoint/database/init.py
```diff
@@ -3,6 +3,8 @@
 
 import sqlalchemy
 
+from bluesearch.entrypoint.database.schemas import schema_articles, schema_sentences
+
```
Stannislav, line 7: Same comment about import as in `add.py`.

tox.ini
```diff
@@ -23,7 +23,9 @@ envlist = lint, py{37, 38, 39}, docs, check-packaging
 
 [testenv]
 download = true
-deps = en-core-web-sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
+deps =
+    en-core-web-sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.0.0/en_core_web_sm-3.0.0.tar.gz
+    en-core-sci-lg @ https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_lg-0.4.0.tar.gz
```
Stannislav, line 28: That's a chonky archive to download for tests (500MB), do you think it would be possible to structure our code so that we can swap `en_core_sci_lg` for `en_core_web_sm` in tests?

pafonta: I guess we could find a code structure allowing to do such swap but it would require much refactoring.

What do you suggest otherwise?

src/bluesearch/entrypoint/database/init.py
pafonta: Done in 5787623.

src/bluesearch/entrypoint/database/add.py
pafonta: Done in 5787623.

src/bluesearch/entrypoint/database/add.py
pafonta: Using `connect()` is not coming from this PR. It is coming from #412.

Maybe @jankrepl could give more context about why #412 didn't use `begin()`?

src/bluesearch/entrypoint/database/add.py
pafonta: Solution in 5787623.

src/bluesearch/entrypoint/database/add.py
pafonta: Nicely spotted. Fixed in 5787623.

src/bluesearch/entrypoint/database/schemas.py
pafonta: This is the length of the MD5 hash produced by `generate_uid(...)`.

pafonta:
Hello @Stannislav,

> I just ran the CLI commands from the description and I'm getting warnings + errors (python3.9):
> ...
> Has anyone had similar issues?

Regarding the warnings:
44a103a should give a solution to this.

Regarding the error:
Is it still the case now? That might be because of the formatting issue from https://github.com/BlueBrain/Search/pull/440#discussion_r709007509.

Stannislav:
> Hello @Stannislav,
> 
> > I just ran the CLI commands from the description and I'm getting warnings + errors (python3.9):
> > ...
> > Has anyone had similar issues?
> 
> Regarding the warnings:
> [44a103a](https://github.com/BlueBrain/Search/commit/44a103af761ba8e9d63d9b663ca6897d2c945b84) should give a solution to this.
> 
> Regarding the error:
> Is it still the case now? That might be because of the formatting issue from [#440 (comment)](https://github.com/BlueBrain/Search/pull/440#discussion_r709007509).

@pafonta thanks, I think it's fixed now:
```
$ bbs_database init test.db
$ bbs_database parse pmc-xml tests/data/sample_file.xml article.pkl
$ bbs_database add test.db article.pkl
$ sqlite3 test.db
sqlite> .tables
articles   sentences
sqlite> select * from articles;
d9ec6b52136a2a3f396ef4a7c43a94f4||||Article Title|Author Given Names 1 Author Surname 1, Author Given Names 2 Author Surname 2|Abstract Paragraph 1
Abstract Paragraph 2||||
sqlite> select * from sentences;
1||Paragraph 1|d9ec6b52136a2a3f396ef4a7c43a94f4|0|0|0
2||Paragraph 2|d9ec6b52136a2a3f396ef4a7c43a94f4|1|0|0
3||Paragraph 3|d9ec6b52136a2a3f396ef4a7c43a94f4|2|0|0
4|Section Title 1|Paragraph Section 1|d9ec6b52136a2a3f396ef4a7c43a94f4|3|0|0
5|Section Title 2|Paragraph Section 2|d9ec6b52136a2a3f396ef4a7c43a94f4|4|0|0
6|Section Title 2|Paragraph 2 Section 2|d9ec6b52136a2a3f396ef4a7c43a94f4|5|0|0
7|Discussion|Paragraph Discussion|d9ec6b52136a2a3f396ef4a7c43a94f4|6|0|0
8|Figure Caption|Figure 1 Caption|d9ec6b52136a2a3f396ef4a7c43a94f4|7|0|0
9|Table Caption|Table 1 Caption|d9ec6b52136a2a3f396ef4a7c43a94f4|8|0|0
10|Table Caption|Table 2 Caption|d9ec6b52136a2a3f396ef4a7c43a94f4|9|0|0
```

pafonta:
Hello @jankrepl,

> Make sure that the 2 tables `articles` and `sentences` are there with the correct columns (after initialization)
> Make sure that the `sentences` table is nonempty (after adding the articles)

Done in b9e515a.

tox.ini
Stannislav: So this model is used in `bluesearch/entrypoint/database/add.py::run`, and I just looked at the unit tests for this and I'm a bit surprised to see that we're calling `main` there and not `run`. I don't thinks it's coming from this PR, must've been introduced at some point earlier.

So, in `test_add.py` instead of 
```python
        args_and_opts = [
            "add",
            engine_sqlite.url.database,
            str(input_path),
            "--db-type=sqlite",
        ]

        main(args_and_opts)
```
I would probably have preferred
```python
run(db_url=engine_sqlite.url.database, path=str(input_path), db_type="sqlite")
```

In this case it would've been easy to parametrize the spacy model by adding a new parameter with a production default to `run`:
```python
def run(
    *,
    db_url: str,
    path: str,
    db_type: str,
    nlp_model_name: str = "en_core_sci_lg",
) -> None: ...
```
and then in unit tests one could put
```python
run(
    db_url=engine_sqlite.url.database,
    path=str(input_path),
    db_type="sqlite",
    nlp_model_name="en_core_web_sm",
)
```

jankrepl: @Stannislav I think that the reason why I decided to go for `main("commandname", ...)` instead of `run_commandname(...)` is to also test that the parsing works. Which is a part of the CLI. However, we can change the logic:) Not sure which one makes more sense.

tests/entrypoint/database/test_init.py
```diff
@@ -25,3 +27,18 @@ def test_sqlite(tmpdir):
...
+
+    expected_metadata = sqlalchemy.MetaData()
+    schema_articles(expected_metadata)
+    schema_sentences(expected_metadata)
+    expected_tables = expected_metadata.sorted_tables
+
+    assert len(tables) == len(expected_tables)
+
+    for table, expected in zip(tables, expected_tables):
+        assert table.compare(expected)
```
jankrepl, line 44: I find this test to be a tautology since it is always going to pass. It feels very similar to:

```
def foo():
     ...

def test_foo():
    assert foo() == foo()
```
IMO we are abusing the fact that we know what is happening inside of `bbs_databse init` (we use the helper `schema_articles` and `schema_sentences`).

Don't you think it would be more useful to just hardcode the column names inside of the tests and then check for set equality?

tox.ini
Stannislav: I see I see. Like you said yourself, maybe it's a bit more in the spirit of unit tests to actually tests the functions separatly.

Like you said, maybe it's possible to write a separate test to check that `get_parser` and `run` match, maybe even without calling `run`, e.g. with the `inspect` module.

jankrepl: Feel free to create an issue:))

tests/entrypoint/database/test_init.py
pafonta: > I find this test to be a tautology since it is always going to pass.

Even with line 33, `metadata.reflect(engine)`?

Indeed, `reflect(...)` loads all available table definitions from the database ([SQLAlchemy docs](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.reflect)).

So, the test really checks that the database has the expected tables and table schemas. Therefore, it doesn't seem that the test does `assert foo() == foo()`.

Or where you thinking about something else?

jankrepl: But how can this test ever fail? We are literally importing and using `schema_articles` and `schema_sentecnes` in the test which is exactly the same thing we are doing in the source code?

pafonta: > But how can this test ever fail?

If `bbs_database init` does not initialize the database? For example, `metadata.create_all` is not called.

The two function `schema_articles` and `schema_sentences` only define the table schemas. They do not initialize the database.

If with this clarifications you still think that's better to check only for the column names and to hardcode them, then it will be done. That will be less maintainable though.

jankrepl: Let's see what @Stannislav @FrancescoCasalegno think:) I am completely willing to make a compromise:) I definitely do see your point of viewing the `schema.py` as a config file rather than source code.

I just think that it would nice to have a failing test if somebody tries to change the schema. Additionally, I think that the asserts inside of unit tests should be as explicit as possible (=enumerating the expected columns) rather

src/bluesearch/entrypoint/database/add.py
Stannislav: @jankrepl do you think we should use `engine.begin()` instead of `engine.connect()`?

tox.ini
Stannislav: Issue: #446

tests/entrypoint/database/test_init.py
Stannislav: Sooo, my 5 cents to put more fuel to the fire 😆.

Maybe it's better to ask: what kind of behaviour are we trying to test here?

If the test hypothesis is "Test that `run` creates a database under given path and populates it with schemas that reflect `schema_articles` and `schema_sentences`" then maybe it's fine.

If, on the other hand, we want to guarantee a certain schema structure then maybe we can have separate tests for `schema_articles` and `schema_sentences`.

jankrepl: Discussed during standup:) Let's keep it the way it is and potentially write a test that really tests the schema at one point in the future.

src/bluesearch/entrypoint/database/add.py
pafonta: As discussed: 2df13d5.

jankrepl (approved):

Stannislav (approved):

tests/integration/test_bbs_database.py
```diff
@@ -70,6 +70,13 @@ def test_bbs_database(tmp_path, setup_backend, jsons_path):
     parsed_files_dir.mkdir()
 
     all_input_paths = sorted(jsons_path.rglob("*.json"))
+
+    # 16e82ce0e0c8a1b36497afc0d4392b4fe21eb174.json and PMC7223769.xml.json are the
+    # same article. Currently, the population of the database is not incremental. So,
+    # the case where the same article is added two times is not handled yet. The patch
+    # below is to move forward until the incremental population is implemented.
+    all_input_paths = [x for x in all_input_paths if x.name != "PMC7223769.xml.json"]
+
```
jankrepl, line 79: Interesting:) I guess there are multiple possible solutions too
* Your solution and potentially remove this `if` statement once we have a real `article_id` generating logic (currently it just uses the article title) and hope that after this the 2 articles are not going to be "identical" anymore
* untrack / remove one of these jsons from our repo
* inplace modify one of the jsons to have , after all this is just test data
* `pytest.raises` when we try to add an identical article the second time around

Anyway, I am totally fine with what you did here:)

jankrepl: btw regarding your comment I would argue that we do have incremental population. However, we just happen to have two articles that are identical (=the same title under the current definition). I guess the fact we get an error when adding an already existing article is actually the desired behavior we agreed on. Or would you disagree?

pafonta: > hope that after this the 2 articles are not going to be "identical" anymore
This is the same article.

> after all this is just test data
I don't know why we have the same article in our test data. Maybe the others could tell us if there is a reason.

So, until then, I suggest to move on.

pafonta: > I would argue that we do have incremental population

The common understanding of "incremental population" implies that duplicates are handled (without crashing).

This is currently not the case as the code crashes (`IntegrityError` from MySQL).

jankrepl (approved):
Looks great:) Thanks!

tests/integration/test_bbs_database.py
pafonta: As discussed:
I have removed the term 'incremental' from the comment in 0e0a84e.

jankrepl: Perfect, thank you very much!:)

pafonta: Regarding the definition of 'incremental population':
The idea is to handle duplicates at the Python code level. This means that `bbs_database add <database url> <directory path of parsed articles>` could run without stopping due to an exception (current situation).

pafonta:
UPDATE

GitHub updated their `ubuntu-latest` images from `20210913.1` to `20210919.1`.

I have noticed that all tests that `pass` are under the version `20210919.1` and most of the tests under `20210913.1` fail.

The fact that ded804c passes but 0e0a84e doesn't is because on 0e0a84e the version `20210913.1` is used.

I think, just clearing the GitHub cache should avoid this `InvocationError`. Unfortunately, we don't have the permission.

PR #442, opened 2021-09-15
Translate relative imports into absolute ones

### LINKED ISSUE #430: Replace relative with absolute imports
Let's improve readability by replacing relative with absolute imports . 

This is recommended in several places.
- [Here](https://stackoverflow.com/a/16748366) — _Python relative imports are no longer strongly discouraged, but using absolute_import is strongly suggested in that case._
- [Here](https://www.python.org/dev/peps/pep-0328/#rationale-for-relative-imports) — _With the shift to absolute imports, the question arose whether relative imports should be allowed at all._

pafonta:
Fixes #430.

## Description

Replace relative with absolute imports.

## How to test?

The following should return nothing:

```
grep -r --exclude-dir "./.*" --include "*.py" "from \." . | sort
```

## Checklist

- [x] All CI tests pass.

### DIFF
--- a/docs/source/whatsnew.rst
+++ b/docs/source/whatsnew.rst
@@ -42,6 +42,7 @@ Latest
 - |Fix| the incorrect maximum input length to the transformer model used as
   backbone for the sentence embedding model :code:`BioBERT NLI+STS CORD-19 v1`.
 - |Add| deterministic generation of paper UIDs based on paper identifiers.
+- |Change| relative imports into absolute ones.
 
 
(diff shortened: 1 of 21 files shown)


jankrepl (approved):
LGTM!!! Thanks:)

Stannislav (approved):
Nice, thanks, very good change :)

FrancescoCasalegno:
Awesome, thanks a lot!

FrancescoCasalegno (approved):

PR #443, opened 2021-09-15
`bbs_database parse` and `add`: add support for multiple files

### LINKED ISSUE #436: Profile `bbs_database parse` and `bbs_database add` + implement `parse_many` and `add_many` (if needed)
With regard to [these results](https://github.com/BlueBrain/Search/pull/412#issuecomment-910356031), profile the commands `bbs_database parse` and `bbs_database add` to see where most of the time is spent.

If this analysis tells that calling several times the same CLI entrypoint is the issue, then implement `parse_many` and `add_many` to be able to parse and add more than one article. In fact, we may even just extend the current interface of `parse` and `add` in a way that they can deal with `>1` article.

FrancescoCasalegno:
Fixes #436 .

## Description

After profiling `bbs_database parse` and `add`, it became clear that running these commands once per each file is not the most efficient approach (see detailed results [here](https://github.com/BlueBrain/Search/issues/436#issuecomment-918981263)).

With this PR, the interface of these two commands is modified to support also running on all the files contained in a given directory. Notice that the output parameter has slightly changed as well.

**Before this PR**
```bash
# Parse a file
bbs_database parse pmc-xml some_article.xml some_article.pkl

# Add a file
bbs_database add database.db some_article.pkl
```

**After this PR**
```bash
# Parse a file / multiple files
bbs_database parse pmc-xml some_article.xml pkl_files/  # parse single file
bbs_database parse pmc-xml xml_files/ pkl_files/        # parse multiple files

# Add a file / multiple files
bbs_database add database.db pkl_files/some_article.pkl  # add single file
bbs_database add database.db pkl_files/                  # add multiple files
```

## Runtime
**Before vs. After this PR**
These are the runtimes obtained by parsing/adding one file at the time, or all files together in bulk.

## How to test?

For a single file:
```bash
bbs_database init tmp.db
bbs_database parse cord19-json tests/data/cord19_v35/document_parses/pmc_json/PMC7140272.xml.json pkl_parses/
bbs_database add tmp.db pkl_parses/PMC7140272.xml.pkl 
```

For multiple files:
```bash
bbs_database init tmp.db
bbs_database parse cord19-json tests/data/cord19_v35/document_parses/pmc_json/ pkl_parses/
bbs_database add tmp.db pkl_parses/
```

## Checklist

- [x] Unit tests added.
  (if needed)
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/entrypoint/database/add.py
+++ b/src/bluesearch/entrypoint/database/add.py
@@ -1,6 +1,24 @@
-"""Adding an article to the database."""
+# Blue Brain Search is a text mining toolbox focused on scientific use cases.
+#
+# Copyright (C) 2020  Blue Brain Project, EPFL.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+"""Adding articles to the database."""
 import argparse
 import pickle  # nosec
+from pathlib import Path
+from typing import Iterable
 
 import sqlalchemy
 
@@ -24,9 +42,9 @@ def get_parser() -> argparse.ArgumentParser:
         """,
     )
     parser.add_argument(
-        "path",
-        type=str,
-        help="""Path to the parsed file.""",
+        "parsed_path",
+        type=Path,
+        help="Path to a parsed file or to a directory of parsed files.",
     )
     parser.add_argument(
         "--db-type",
@@ -41,7 +59,7 @@ def get_parser() -> argparse.ArgumentParser:
 def run(
     *,
     db_url: str,
-    path: str,
+    parsed_path: Path,
     db_type: str,
 ) -> None:
     """Add an entry to the database.
@@ -59,9 +77,22 @@ def run(
         # This branch never reached because of `choices` in `argparse`
         raise ValueError(f"Unrecognized database type {db_type}.")  # pragma: nocover
 
-    with open(path, "rb") as f:
-        article = pickle.load(f)  # nosec
+    inputs: Iterable[Path]
+    if parsed_path.is_file():
+        inputs = [parsed_path]
+    elif parsed_path.is_dir():
+        inputs = parsed_path.glob("*")
+    else:
+        raise ValueError(
+            "Argument 'parsed_path' should be a path to an existing file or directory!"
+        )
+
+    titles = []
+    for inp in inputs:
+        with inp.open("rb") as f:
+            article = pickle.load(f)  # nosec
+        titles.append({"title": article.title})
 
     with engine.connect() as con:
         query = sqlalchemy.text("INSERT INTO articles(title) VALUES(:title)")
-        con.execute(query, {"title": article.title})
+        con.execute(query, *titles)
--- a/src/bluesearch/entrypoint/database/add.py
+++ b/src/bluesearch/entrypoint/database/add.py
@@ -81,7 +81,7 @@ def run(
     if parsed_path.is_file():
         inputs = [parsed_path]
     elif parsed_path.is_dir():
-        inputs = parsed_path.glob("*")
+        inputs = sorted(parsed_path.glob("*"))
     else:
         raise ValueError(
             "Argument 'parsed_path' should be a path to an existing file or directory!"
--- a/src/bluesearch/entrypoint/database/add.py
+++ b/src/bluesearch/entrypoint/database/add.py
@@ -81,7 +81,7 @@ def run(
     if parsed_path.is_file():
         inputs = [parsed_path]
     elif parsed_path.is_dir():
-        inputs = sorted(parsed_path.glob("*"))
+        inputs = sorted(parsed_path.glob("*.pkl"))
     else:
         raise ValueError(
             "Argument 'parsed_path' should be a path to an existing file or directory!"
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -86,7 +86,7 @@ def run(
     for inp in inputs:
         parser_inst: ArticleParser
         if article_type == "cord19-json":
-            with inp.open("r") as f_inp:
+            with inp.open() as f_inp:
                 parser_inst = CORD19ArticleParser(json.load(f_inp))
         elif article_type == "pmc-xml":
             parser_inst = PubmedXMLParser(inp)
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -18,6 +18,7 @@
 import argparse
 import json
 import pickle  # nosec
+import warnings
 from pathlib import Path
 from typing import Iterable
 
@@ -83,20 +84,26 @@ def run(
             "Argument 'input_path' should be a path to an existing file or directory!"
         )
 
+    output_path.mkdir(exist_ok=True)
+
     for inp in inputs:
-        parser_inst: ArticleParser
-        if article_type == "cord19-json":
-            with inp.open() as f_inp:
-                parser_inst = CORD19ArticleParser(json.load(f_inp))
-        elif article_type == "pmc-xml":
-            parser_inst = PubmedXMLParser(inp)
-        else:
-            raise ValueError(f"Unsupported article type {article_type}")
+        try:
+            parser_inst: ArticleParser
+            if article_type == "cord19-json":
+                with inp.open() as f_inp:
+                    parser_inst = CORD19ArticleParser(json.load(f_inp))
+            elif article_type == "pmc-xml":
+                parser_inst = PubmedXMLParser(inp)
+            else:
+                raise ValueError(f"Unsupported article type {article_type}")
 
-        article = Article.parse(parser_inst)
+            article = Article.parse(parser_inst)
 
-        output_path.mkdir(exist_ok=True)
-        out = (output_path / inp.name).with_suffix(".pkl")
+            out = (output_path / inp.name).with_suffix(".pkl")
+            with out.open("wb") as f_out:
+                pickle.dump(article, f_out)
 
-        with out.open("wb") as f_out:
-            pickle.dump(article, f_out)
+        except Exception as e:
+            warnings.warn(
+                f'Failed parsing file "{inp}":\n {e}', category=RuntimeWarning
+            )
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -96,8 +96,7 @@ def run(
         article = Article.parse(parser_inst)
 
         output_path.mkdir(exist_ok=True)
-        # If we used .with_suffix(), then we would get file.pdf.xml --> file.pkl
-        out = output_path / (inp.stem + ".pkl")
+        out = (output_path / inp.name).with_suffix(".pkl")
 
         with out.open("wb") as f_out:
             pickle.dump(article, f_out)
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -77,7 +77,7 @@ def run(
     if input_path.is_file():
         inputs = [input_path]
     elif input_path.is_dir():
-        inputs = input_path.glob("*")
+        inputs = sorted(input_path.glob("*"))
     else:
         raise ValueError(
             "Argument 'input_path' should be a path to an existing file or directory!"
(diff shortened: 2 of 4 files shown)


src/bluesearch/entrypoint/database/parse.py
```diff
@@ -14,49 +32,72 @@
...
+        inputs = input_path.glob("*")
     else:
-        raise ValueError(f"Unsupported article type {article_type}")
+        raise ValueError(
+            "Argument 'input_path' should be a path to an existing file or directory!"
+        )
+
+    for inp in inputs:
+        parser_inst: ArticleParser
+        if article_type == "cord19-json":
```
Stannislav, line 88: In principle we could do this if once outside of the for-loop instead of at every iteration.

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -14,49 +32,72 @@
...
+            with inp.open("r") as f_inp:
```
Stannislav, line 89: I think `"r"` is the default parameter anyway and can be dropped.

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -14,49 +32,72 @@
...
+        elif article_type == "pmc-xml":
+            parser_inst = PubmedXMLParser(inp)
+        else:
+            raise ValueError(f"Unsupported article type {article_type}")
+
+        article = Article.parse(parser_inst)
 
-    article = Article.parse(parser_inst)
+        output_path.mkdir(exist_ok=True)
+        # If we used .with_suffix(), then we would get file.pdf.xml --> file.pkl
```
Stannislav, line 99: I'm getting a different result 🤔 
```python
>>> p = pathlib.Path("a.pdf.xml")
>>> p.with_suffix(".pkl")
PosixPath('a.pdf.pkl')
```

tests/entrypoint/database/test_add.py
```diff
@@ -36,34 +35,60 @@ def test_mysql_not_implemented():
...
+            str(pkl_file),
+            "--db-type=sqlite",
+        ]
+        main(args_and_opts)
+    (n_rows,) = engine_sqlite.execute(query).fetchone()
+    assert n_rows == n_files
+
+    engine_sqlite.execute("DELETE FROM articles WHERE True")
+    (n_rows,) = engine_sqlite.execute(query).fetchone()
+    assert n_rows == 0
```
Stannislav, line 73: What does this do exactly? I'm not familar with `WHERE True`.

src/bluesearch/entrypoint/database/add.py
```diff
@@ -59,9 +77,22 @@ def run(
         # This branch never reached because of `choices` in `argparse`
         raise ValueError(f"Unrecognized database type {db_type}.")  # pragma: nocover
 
-    with open(path, "rb") as f:
-        article = pickle.load(f)  # nosec
+    inputs: Iterable[Path]
+    if parsed_path.is_file():
+        inputs = [parsed_path]
+    elif parsed_path.is_dir():
+        inputs = parsed_path.glob("*")
```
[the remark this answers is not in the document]
FrancescoCasalegno: I do like your idea, but I thought that this would set a precedent for the similar `glob("*")` in `parse.py`.

And in `parse.py` it is much harder to guess the extension. 
- First, because the extension depends on the format — e.g. `*.json` for `cord19-json` and `*.xml` for `pmc-xml`.
- Second, because even for the same format you may have different extensions — e.g. `*.xml` and `*.nxml` are both used for `pmc-xml` files).

src/bluesearch/entrypoint/database/add.py
```diff
@@ -24,9 +42,9 @@ def get_parser() -> argparse.ArgumentParser:
         """,
     )
     parser.add_argument(
-        "path",
-        type=str,
-        help="""Path to the parsed file.""",
+        "parsed_path",
+        type=Path,
```
jankrepl, line 46: Wow, this actually makes sense:) Did not realize that this is possible:)

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: I thought about it , but I saw two issues.
- With the approach that we took (1 article : 1 parser object) we need to create a dedicated `parser_inst = CORD19ArticleParser(json.load(f_inp))` for each `f_inp` file anyway.
- I thought about using a `dict` to implement a factory outside of the loop, but since the parsers don't have a common interface (see #419) this becomes really complicated...

If you think we can still move it outside the loop, do you have a specific suggestion on how to do it?

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: It is indeed the default, I just added it there for more clarity.

Removed in 3baf7c2 .

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -14,49 +32,72 @@
...
-        with open(input_path) as f_input:
-            parser_inst = CORD19ArticleParser(json.load(f_input))
-    elif article_type == "pmc-xml":
-        parser_inst = PubmedXMLParser(input_path)
-
+    inputs: Iterable[Path]
+    if input_path.is_file():
+        inputs = [input_path]
+    elif input_path.is_dir():
+        inputs = input_path.glob("*")
```
jankrepl, line 80: Does the `glob(*)` have some sorting logic?

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: I was referring to the following line, where `.stem` is used. 

I did not k(now until I saw a test failing) that `.stem` removes _all_ extensions from a given file path. I think this is not desirable, since the previous extensions could actually carry information on the article, and would help identifying the source in a unique way.

The following snipped hopefully clarifies what I mean.
```python
>>> fname = Path("neuroscience_article.xml.json")
>>> out_dir = Path("out_pkls/")
>>> (out_dir / fname.stem).with_suffix(".pkl")
PosixPath('out_pkls/neuroscience_article.pkl')
>>> out_dir / (fname.stem + ".pkl")
PosixPath('out_pkls/neuroscience_article.xml.pkl')
```

tests/entrypoint/database/test_add.py
FrancescoCasalegno: It just drops all rows in the table.

`WHERE True` is not needed per se, I just put it there to make explicit my intent of deleting _all_ rows in the table. See e.g. StackOverflow [here](https://stackoverflow.com/a/37849060).

For some reason, `DELETE FROM articles` sounds a bit awkward to me if there is no `WHERE ...` clause following it 😁 
But if you find it confusing I can change it!

jankrepl (commented):
Looks great!

One think I would like to discuss is the behavior when there is a corrupted file inside of a folder (anything that will lead to Python crashing)
* The `parse` command is dumping pickles inside of a for loop. Unless the corrupted file is the first one in line the `parse` command will generate some pickle files even if it exits with a nonzero code. If the order in which the files are parsed is random then this could even lead to different results over different runs.
* The `add` command behaves differently since it loads the pickles into a Python list and then sends them to the database all at once.

src/bluesearch/entrypoint/database/parse.py
jankrepl: BTW flake8 or black are not catching this, right? I honestly do not mind seeing the default parameters being rewritten in function calls. Especially when they are important.

Also, I can argue that if the 3rd party package changes their defaults, then we might get into trouble:D

src/bluesearch/entrypoint/database/parse.py
Stannislav: I'm still not sure `Path.stem` removes all extensions:
```python
>>> fname = Path("neuroscience_article.xml.json")
>>> fname.stem
'neuroscience_article.xml'
```
I think it's just because in your example `(out_dir / fname.stem) = "out_pkls/neuroscience_article.xml"` and the subsequent `with_suffix` replaces `.xml` by `.pkl`.

For example this would work:
```python
>>> inp
PosixPath('/some/dir/neuroscience_article.xml.json')
>>> out_dir
PosixPath('out_pkls')
>>> out_dir / inp.with_suffix(".pkl").name
PosixPath('out_pkls/neuroscience_article.xml.pkl')
```

tests/entrypoint/database/test_add.py
```diff
@@ -9,8 +8,8 @@
 
 
 @pytest.fixture()
-def engine_sqlite(tmpdir):
-    db_url = pathlib.Path(str(tmpdir)) / "database.db"
+def engine_sqlite(tmp_path):
```
jankrepl, line 11: haha, how did I not know about `tmp_path`? We should probably use it everywhere in our codebase instead of `pathlib.Path(str(tmpdir))`.

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: You're right! Thanks for the clarification, fixed in b6bdb08.

tests/entrypoint/database/test_add.py
FrancescoCasalegno: Changed in 50257b0.

src/bluesearch/entrypoint/database/parse.py
FrancescoCasalegno: As you correctly said, `Path.glob()` does not guarantee any specific order.
Fixed in ae81762 to guarantee reproducibility.

tests/entrypoint/database/test_add.py
FrancescoCasalegno: I like `tmp_path` because it saves an `import` statement (and hopefully some time as a consequence) 😄

tests/entrypoint/database/test_add.py
Stannislav: I see, thanks for the explanation! :)

FrancescoCasalegno:
@jankrepl 

> One thing I would like to discuss is the behavior when there is a corrupted file inside of a folder (anything that will lead to Python crashing)

You are making a very good comment here about the behavior of `parse` and `add`! If I correctly understand, you are saying that, as of now, `add` is **deterministic** + [**atomic**](https://en.wikipedia.org/wiki/Atomicity_(database_systems)) while `parse` is **non-deterministic** + non-atomic.

If the issue with `parse` is just about non-determinism, this should be fixed in ae81762. But if you would like `parse` to also be atomic, I think the solution is not so trivial. What I can do is to first load all the parsed articles in a list, and then dump them one by one in the `pickle` format. But:
1. One of the `pickle.dump()` could fail in the middle of the list (unlikely, but you never know), so that the operation still wouldn't be atomic. 
To make this operation truly atomic, a solution would be "keeping a copy of the data before any changes occurred" ([source](https://en.wikipedia.org/wiki/Atomicity_(database_systems)#Implementation)), but this sounds quite costly and still not trivial to implement.
2. Loading all the parsed articles in a list could make memory blow up. However, I agree that this is already pretty much what we are doing in `add`.
We could improve the memory impact by running `parse` and `add` by batches of user-defined size. But then we would lose the atomicity for both operations...

What do you suggest to do?

src/bluesearch/entrypoint/database/parse.py
Stannislav: Ah yes, you're totally right, I don't think there's a way to avoid an if-statement inside the for-loop. My bad!

src/bluesearch/entrypoint/database/add.py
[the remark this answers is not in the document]
Stannislav: I see the problme for `parse`, but here we'll always have one single fixed extension, right?

Regarding `parse`, there we can actually run into the same trouble of having random foreign files being around and crashing our parsers. I don't see a super straightforward solution for `parse` except catching potential exceptions... Could be a separate issue.

jankrepl:
> You are making a very good comment here about the behavior of parse and add! If I correctly understand, you are saying that, as of now, add is deterministic + atomic while parse is non-deterministic + non-atomic.

Yeh, exactly!

> If the issue with parse is just about non-determinism, this should be fixed in ae81762. But if you would like parse to also be atomic, I think the solution is not so trivial. 

I agree.


I guess I am fine with `add` being atomic and `parse` non-atomic. However, I have to specific notes:
1. Why not have a `try ... except` block inside of `parse` so that we can skip to the next file if the current one was impossible to parse? Rather than stopping the entire process. For example, if we want to parse a folder of 100 files where 2 are unparsable then we would be guaranteed to end up with 98 parsed files whereas now it completely depends on the order. Note that ae81762 made the order deterministic, but still does not guarantee that we parse as many files as possible.
2. If we go for the above, then we probably should have a mechanism to let the user know which files failed to be parsed. Some ideas are logging, warnings or just plain standard output. Related to this, we will probably use `bbs_database parse` in shell scripts and therefore we might want to consider what exit codes we assign to the 3 possible scenarios.
   1. All files were parsed
   2. Some files were parsed (probably one or more files were unparsable)
   3. No file was parsed (potentially problem within the Python implementaiton or wrong usage of the entrypoint)


Anyway, we can totally turn this comment into an issue and we can deal with it in a different PR or not address it at all:)

src/bluesearch/entrypoint/database/add.py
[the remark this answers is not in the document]
FrancescoCasalegno: Done in 10e2e28.
In addition to that, I created a dedicated issue https://github.com/BlueBrain/Search/issues/445 for `bbs_database parse` and I tried to also provide some possible solutions.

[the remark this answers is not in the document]
Stannislav: awesome, thanks!

Stannislav (approved):

FrancescoCasalegno:
@jankrepl 
> Why not have a `try ... except` block inside of parse [...] we probably should have a mechanism to let the user know which files failed [...]

Should be done in 3910bfb. I am printing a `RuntimeWarning` for any file that failed to parse.

jankrepl (approved):

PR #455, opened 2021-09-23
#448 Create unique article_id column when bbs_database add is called

### LINKED ISSUE #448: Create unique `article_id` column when `bbs_database add` is called
## 🚀 Feature

In https://github.com/BlueBrain/Search/pull/418 we introduced utilities to compute a unique `article_id` for each article ingested into our database. 

We now need to integrate this into `bbs_database add` command, so that the `article` table will have a `PRIMARY KEY` (and therefore `UNIQUE`, `NOT NULL`) column `article_id` filled as new papers are ingested.

This implies also extracting identifiers when parsing the article. See also [these lines](https://github.com/BlueBrain/Search/blob/b9e515a63592e992e31420f99b5424a703831909/src/bluesearch/entrypoint/database/add.py#L71-L72) from https://github.com/BlueBrain/Search/pull/440.

EmilieDel:
Fixes #448.

## Description

This PR handles the concrete parsing of articles identifiers (currently namely: `(PubMed ID, PMC ID and DOI)`) and the generation of the unique ID. 
- All of them can be found in the schema of `XML` files of PMC papers. 
- Only `PMC` id is specified concerning the `json` files of CORD-19.
- From PMC papers on DGX1:
    - All of them possess a PMC ID.
    - PubMed ID is specified in 89.85 %.
    - DOI is specified in 83.51 %.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/article.py
+++ b/src/bluesearch/database/article.py
@@ -21,7 +21,7 @@ import html
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Generator, Iterable, Sequence
+from typing import Generator, Iterable, Optional, Sequence
 from xml.etree.ElementTree import Element  # nosec
 
 from defusedxml import ElementTree
@@ -518,6 +518,10 @@ class Article:
     authors: Sequence[str]
     abstract: Sequence[str]
     section_paragraphs: Sequence[tuple[str, str]]
+    pubmed_id: Optional[str]
+    pmc_id: Optional[str]
+    doi: Optional[str]
+    uid: Optional[str]
 
     @classmethod
     def parse(cls, parser: ArticleParser) -> Article:
@@ -532,8 +536,14 @@ class Article:
         authors = tuple(parser.authors)
         abstract = tuple(parser.abstract)
         section_paragraphs = tuple(parser.paragraphs)
+        pubmed_id = parser.pubmed_id
+        pmc_id = parser.pmc_id
+        doi = parser.doi
+        uid = parser.uid
 
-        return cls(title, authors, abstract, section_paragraphs)
+        return cls(
+            title, authors, abstract, section_paragraphs, pubmed_id, pmc_id, doi, uid
+        )
 
     def iter_paragraphs(
         self, with_abstract: bool = False
--- a/src/bluesearch/database/article.py
+++ b/src/bluesearch/database/article.py
@@ -78,7 +78,7 @@ class ArticleParser(ABC):
         """
 
     @property
-    def pubmed_id(self) -> str | None:
+    def pubmed_id(self) -> Optional[str]:
         """Get Pubmed ID.
 
         Returns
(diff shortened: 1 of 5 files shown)


src/bluesearch/database/article.py
```diff
@@ -75,6 +77,51 @@ def paragraphs(self) -> Iterable[tuple[str, str]]:
             is the section title, the second the paragraph content.
         """
 
+    @property
+    def pubmed_id(self) -> str | None:
```
jankrepl, line 81: I guess this might be just a matter of convention, but should we maybe stick to a single convention?

* https://stackoverflow.com/a/51710151 
* https://www.python.org/dev/peps/pep-0604/

 All the below should be equivalent:

* `Optional[str]`
* `str | None`
* `Union[str, None]`

I guess `Optional[str]` is something we have used so far, however, as per PER 604`str | None` might be the recommended way in new Pythons.

Anyway, I am not requiring any actions here other than just getting everybody's opinion:) @EmilieDel @pafonta

src/bluesearch/database/article.py
```diff
@@ -393,6 +494,17 @@ def paragraphs(self) -> Generator[tuple[str, str], None, None]:
...
+    @property
+    def pmc_id(self) -> str | None:
+        """Get PMC ID.
+
+        Returns
+        -------
+        str or None
+            PMC ID if specified, otherwise None.
+        """
+        return self.data["paper_id"]
```
jankrepl, line 506: Shouldn't we do `self.data.get("paper_id")` in case it is not there?

jankrepl (approved):
Looks great! Thank you!

src/bluesearch/database/article.py
EmilieDel: You are right, I forgot to change it! Thanks for catching this. Done in 76fc0ef.

src/bluesearch/database/article.py
EmilieDel: I agree about sticking to a single convention. I switched to `Optional[str]` convention as I think it is the most used one for now. If we consider changing this for `str | None` maybe we can create an issue and change it for the entire package ?

src/bluesearch/database/article.py
```diff
@@ -189,6 +237,59 @@ def paragraphs(self) -> Generator[tuple[str, str], None, None]:
...
+        ids : dict
+            Dictionary whose keys are ids type and value are ids values.
+        """
+        ids = {}
+        article_ids = self.content.findall("./front/article-meta/article-id")
+
+        for article_id in article_ids:
+
+            if "pub-id-type" not in article_id.attrib.keys():
+                continue
```
pafonta, line 287: In which context would this case happen?

pafonta (approved):
Good to have now the identifiers parsed.

src/bluesearch/database/article.py
EmilieDel: I am not about a concrete explanation for this, but I had the issue once.

PR #457, opened 2021-09-27
#439 Use topic info when populating database

### LINKED ISSUE #439: Use topic info when populating database
## Context

- In #432 we added utilities for retrieving topic info about a journal or an article.
- However, these utilities are not integrated yet anywhere in our pipeline, so the topic info is not used.
- We want to filter papers based on this topic info.

## Possible solutions

I can mainly think of two possible approaches. I tried to summarize the main pros and cons that I can see here.
1. **Filter papers before/while populating the database.**
   - **Pros**
       - Easy implementation — e.g. just add a check in `bbs_database add`.
       - Reduced database size — e.g. avoid having PMC articles from "Windows 10 for the Internet of Things" in a NeuroSci db.
    - **Cons**
       - Discard useful papers —  e.g. a paper w/o topic keywords (or with wrong keywords) may be discarded and we won't be able to change our mind during search/mining, we just won't have them in our database.
       - No dynamic filtering — e.g. user cannot decide to consider more/less/other filtering keywords at runtime.
2. **Filter papers at search/mining time.**
    - **Pros**
       - Dynamic filtering.
    - **Cons**
       - More complex implementation – e.g. we would need to store topic info on each article in `article` table, and info on each journal in `journal` table (but notice that having this topic info stored also allow us to update it at any time).
       - Larger database size

**NOTE**. We could also support both approaches.

EmilieDel:
Fixes #439.

## Description

## How to test?

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.


src/bluesearch/database/topic.py
```diff
@@ -27,13 +27,13 @@
 
 
 # Journal Topic
-def request_mesh_from_nlm_ta(nlm_ta: str) -> list[dict] | None:
-    """Retrieve Medical Subject Heading from Journal's NLM Title Abbreviation.
+def request_mesh_from_journal_title(journal_title: str) -> list[dict] | None:
+    """Retrieve Medical Subject Heading from Journal's NLM Title.
```
pafonta, line 31: Then, is there at the end a difference between using the abbreviated title and the full title?

tests/unit/database/test_topic.py
```diff
@@ -37,7 +37,8 @@ def test_normal_behaviour_works(self, test_data_path):
             responses.GET,
             (
                 "https://www.ncbi.nlm.nih.gov/nlmcatalog?"
-                "term=Trauma Surg Acute Care Open[ta]&report=xml&format=text"
```
pafonta, line 40: Is passing from `[ta]` to `[Journal]` the only change needed to use full titles?

src/bluesearch/database/topic.py
pafonta: By the way, why the change?

tests/unit/database/test_topic.py
EmilieDel: Yes

src/bluesearch/database/topic.py
EmilieDel: (Just a side note, It is still a draft PR)
- Initially, when searching at `how to find the topic of an article`, I based the research on the `pmc` folder structure which is:
```
nlm_abbreviated_title/
    PMCXXX.nxml
    ....
```
So it was making sense to look at the journal thanks to the abbreviated title.
-  Thinking at a more generic strategy (we won't always parse PMC articles coming directly from the mentioned structure), I parsed the PMC files we have to see which info was communicated about the journal. The title seems more often specified than the journal abbreviated title. 
This is the reason for the change. But this is a draft PR, I am still considering staying with the search of abbreviated title or even have the two possibilities in the source code.

PR #458, opened 2021-09-27
Add JSON serialization for dataclass Article

### LINKED ISSUE #422: Do not use pickle to store parsed articles
The PR #412 introduced (among other things) a parsing logic (`bbs_database parse`) that ends up writing a pickle file.

However, this is not optimal. A  few reasons:
* Security risk since there is code execution when loading (we even explicitly ignore the `bandit` warning)
* If the `Article` source code changes we might not be able to load existing pickle files

We should look for alternatives.

pafonta:
Fixes #422.

## Description

Between parsing and loading, the data of `Article` instances are serialized as `pickle` files. This implements the best practice named 'separation of concerns'. Indeed, the logic for parsing raw data into `Article` instances and the logic for loading `Article` instances in the database are then separated.

The issue #422 wants to move away from the `pickle` format. Indeed, the tool `bandit` warns against it. If someone would have added malicious code to the serialized `Article` instances, deserializing the data could lead to execute malicious code.

➡️ The goal of the current PR is then to replace the `pickle` format with another format. This other format is `JSON`.

A Python library named `mashumaro` ([PyPI](https://pypi.org/project/mashumaro/)), lets serialize and deserialize a `dataclass` into several formats. One is `JSON`.

The type annotation `tuple[]` was replaced with `Tuple[]`. Indeed, we use a bit of this Python 3.9+ style for typing (`tuple[str, str]`) thanks to a `__future__` import. However, `mashumaro` uses this Python 3.9+ style only on Python 3.9+.

## How to test?

The changes are automatically tested by these tests:
- `tests/unit/entrypoint/database/test_parse.py`
- `tests/unit/entrypoint/database/test_add.py`

## Checklist

- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/article.py
+++ b/src/bluesearch/database/article.py
@@ -21,10 +21,11 @@ import html
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Generator, Iterable, Sequence
+from typing import Generator, Iterable, List, Tuple
 from xml.etree.ElementTree import Element  # nosec
 
 from defusedxml import ElementTree
+from serde import deserialize, serialize
 
 
 class ArticleParser(ABC):
@@ -398,14 +399,16 @@ class CORD19ArticleParser(ArticleParser):
         return f'CORD-19 article ID={self.data["paper_id"]}'
 
 
+@deserialize
+@serialize
 @dataclass(frozen=True)
 class Article:
     """Abstraction of a scientific article and its contents."""
 
     title: str
-    authors: Sequence[str]
-    abstract: Sequence[str]
-    section_paragraphs: Sequence[tuple[str, str]]
+    authors: List[str]
+    abstract: List[str]
+    section_paragraphs: List[Tuple[str, str]]
 
     @classmethod
     def parse(cls, parser: ArticleParser) -> Article:
@@ -417,9 +420,9 @@ class Article:
             An article parser instance.
         """
         title = parser.title
-        authors = tuple(parser.authors)
-        abstract = tuple(parser.abstract)
-        section_paragraphs = tuple(parser.paragraphs)
+        authors = list(parser.authors)
+        abstract = list(parser.abstract)
+        section_paragraphs = list(parser.paragraphs)
 
         return cls(title, authors, abstract, section_paragraphs)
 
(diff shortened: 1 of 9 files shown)


jankrepl (approved):
Looks good! Interesting approach :) Thank you

src/bluesearch/database/article.py
```diff
@@ -510,14 +511,16 @@ def __str__(self):
...
 class Article:
     """Abstraction of a scientific article and its contents."""
 
     title: str
-    authors: Sequence[str]
-    abstract: Sequence[str]
-    section_paragraphs: Sequence[tuple[str, str]]
+    authors: List[str]
+    abstract: List[str]
+    section_paragraphs: List[Tuple[str, str]]
```
EmilieDel, line 523: Is it not possible to switch to `Tuple` here instead of `List` ?

EmilieDel (approved):
Looks good to me.

src/bluesearch/database/article.py
pafonta: That's not what happened.

We changed:
1. for `List` instead of `Sequence`,
2. and for `Tuple` instead of `tuple`.

The (1) is because `pyserde` does not support `Sequence` yet but supports `List`.
The (2) is because `pyserde` does not support the notation `tuple` from Python 3.9+ for earlier Python versions.

EmilieDel: My question is more, when not changing `Sequence` to `Tuple` ?

pafonta: No typing change is required any more with the new library: `mashumaro`.

I have updated the description of the PR.

pafonta: Following our discussion, I stumble into the fact that `pyserde` doesn't support tuples of variable length.

I went then for another library named `mashumaro`. It is said to be a bit slower. However, it has built-in support for `Sequence`.

EmilieDel: Thanks a lot for looking at this and changing for `Sequence`. Would you mind adding a comment in `setup.py` next to `mashumaro` (the name is not very explicit) so that we know we use this package for the JSON serialization ? :)

pafonta: Sure. Done in 0fc157f.

EmilieDel: Sounds good for me! Thanks a lot.

PR #495, opened 2021-11-03
 Split up the CI jobs into smaller jobs

### LINKED ISSUE #494: Split up the CI jobs into smaller jobs
## 🚀 Feature
Create new smaller jobs in our CI GitHub workflow - one for linting, unit tests, docs, twine...

## Motivation
* No need to run linting/checks in all python versions
* Things will run in parallel instead of sequentially
* Better visibility on what's actually failing

Stannislav:
Fixes #494 .

## Description

Create new smaller jobs in our CI GitHub workflow - one for linting, unit tests, docs, twine...

Motivation:
* No need to run linting/checks in all python versions
* Things will run in parallel instead of sequentially
* Better visibility on what's actually failing

Also start caching pip wheels.

## How to test?

Check the entries in the GitHub actions runs in this PR

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
...
+        tox-env: [lint, docs, check-packaging]
+    steps:
+      - name: Cancel previous workflows that are still running
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+      - name: Checkout latest commit
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # fetch all history with version tags
```
pafonta, line 25: What is our use case for this new element?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: Set up pip cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ matrix.tox-env }}-${{ runner.os }}-${{ hashFiles('tox.ini') }}
```
pafonta, line 34: Is it useful? Indeed, with the current key, the cache will never be reused because `matrix.tox-env` is unique. However, I am new to https://github.com/actions/cache and I might have missed something in the documentation.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
...
+          fetch-depth: 0  # fetch all history with version tags
+      - name: Set up python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Set up pip cache
+        uses: actions/cache@v2
+        with:
+          path: .cache/pip
+          key: ${{ matrix.tox-env }}-${{ matrix.os }}-${{ hashFiles('tox.ini') }}
```
pafonta, line 74: Same comment as above about utility with the current `key`.

pafonta (commented):
Thanks for organizing the tests more.

I have a question and some review comments.

Is there a way to avoid repeating the common workflow (cancel-workflow-action, checkout, setup-python, cache) between jobs (tox, unit-tests)?

.github/workflows/ci.yaml
Stannislav: I think the idea is not to share the cache between the jobs, but to re-use it the next time the same job is run.

.github/workflows/ci.yaml
Stannislav: Yes, thanks from bringing this up.

The idea is that we use `setuptools-scm` to generate our version from git tags, but with the default shallow clone there will be not tags, and so the generated version will always be some default (`0.1` I believe).

I don't think this causes any problems for us, I saw this being a problem elsewhere, where the tests relied on the presence of the patch version, which is otherwise missing in the default version.

So all-in-all, for us this is not strictly necessary. On the other hand this step takes 2-3 seconds, so not a big deal either. What do you guys think?

Stannislav:
> Is there a way to avoid repeating the common workflow (cancel-workflow-action, checkout, setup-python, cache) between jobs (tox, unit-tests)?

I'm wondering about the same. I suspect there must be, but I don't know how to do this off the top of my head. Also didn't have time to research this. Do you have any insights? There's probably a way to define a template set of steps.

.github/workflows/ci.yaml
pafonta: I see. Resolving.

Note: We need then to remember this on long running pull requests. Indeed, we might have older package versions on such pull request compared to a more recent one, while the CI runs are done around the same time.

.github/workflows/ci.yaml
pafonta: Resolved in https://github.com/BlueBrain/Search/pull/495#discussion_r742619196.

.github/workflows/ci.yaml
pafonta: Fair enough. OK to have this new configuration. Resolving.

.github/workflows/ci.yaml
Stannislav: Yes, good point. Would probably good to clear the cache if we change our dependencies drastically.

pafonta (approved):
Thanks for addressing the review comments.

> > Is there a way to avoid repeating the common workflow (cancel-workflow-action, checkout, setup-python, cache) between jobs (tox, unit-tests)?
> 
> I'm wondering about the same. I suspect there must be, but I don't know how to do this off the top of my head. Also didn't have time to research this. Do you have any insights? There's probably a way to define a template set of steps.

I think this is unusual to separate the jobs between `tox` and `pytest`, like introduced by the PR. So, I haven't a reference on creating templates of steps in my head.

Maybe one could create an action regrouping the common actions between the steps. That would achieve the same goal as a template of steps. But that might be overkilled at this point.

Stannislav:
> That would achieve the same goal as a template of steps. But that might be overkilled at this point.

Yeah, I'm more gravitating towards leaving it for now. I doulbe checked and there are really only three steps that are completely identical (cancel previous, checkout, set up environment) and they're not even consecutive. So I'm not sure it's worth the time working this out.

jankrepl (approved):
Looks good I guess:) And I seems to be faster than `master`.

Stannislav:
I think I'm going ahead and merge this one now since it's got 2 approvals. Hopefullyl it's fine with everyone :)

Issue #429, opened 2021-08-24
Handle empty paragraphs/fields in PubmedXMLParser


EmilieDel:
Currently, empty paragraphs/fields are kept by the parser PubmedXMLParser. (see discussion and comments in #406)

It could be nice to:
[ ] Analyse if some papers have empty paragraphs/fields when we parse them
[ ] Decide what to do in that case

Issue #425, opened 2021-08-24
Replace double-slash syntax of PubmedXMLParser


EmilieDel:
## 🚀 Feature

Currently, we use `element.find(".//some/path")` syntax into the `PubmedXMLparser`, [the double-slash is a glob for all elements at all sub-levels](https://docs.python.org/3/library/xml.etree.elementtree.html#xpath-support). If we know the exact (fixed) structure of documents, we might consider using a single slash instead. Might be worth checking the performance impact.

_Comment initially made by @Stannislav in [#406 (comment)](https://github.com/BlueBrain/Search/pull/406#pullrequestreview-737356877)

[ ] Compare performance impact in terms of runtime between double-slash and single one syntax 
[ ] Check if replacing double-slash by single-slash syntax is working for different version of DTD JATS schema
[ ] If it working for all versions and helps in terms of performance, replace double-slash by single-slash syntax

Issue #449, opened 2021-09-16
Parse lists and tables in XML files correctly


Stannislav:
In #437 a custom parsing of XML tags was introduced. At the moment only a subset of tags are properly processed, the rest go through a default fall-back processing which just extracts all the text.

While the default handling is good enough for most cases it appears that other tags like `list` and `table-wrap` need to be handled in a better way.

# To do
- Identify tags for which the processing needs to be imporved (see the big table in #437)
- Cross-check on the real data from the open-access dataset
- Implement a handling for those tags
- Post the updated tag implementation table here.

# Details
Another 2 tags that I think are important but at the same time relatively feasible to handle are `list` and `table-wrap`. For `list` in particular, I think we can pretty much just extract the text, but maybe we need to double check to make sure nothing weird happens.

Maybe you can create one issue for each of those two?

_Originally posted by @FrancescoCasalegno in https://github.com/BlueBrain/Search/pull/437#discussion_r710043549_

pafonta:
⚠️  Currently, the elements of lists are duplicated by the parser (`PMCXMLParser`)!

➡️  The literature database contains duplicated sentences from this parsing issue.

<details>
  <summary>details</summary>

Example XML of an article with a list in a paragraph:

```xml
<article>
    <body>
        <p>
            <list list-type="bullet">
                <list-item>
                    <p>List item 1.</p>
                </list-item>
                <list-item>
                    <p>List item 2.</p>
                </list-item>
            </list>
        </p>
    </body>
</article>
```

Parsing of such article file (article.xml):

```python
from bluesearch.database.article import PMCXMLParser

parser = PMCXMLParser("article.xml")
for paragraph in parser.paragraphs:
  print(paragraph)
```

Duplicated outputs:

```
('', 'List item 1.\n                List item 2.')
('', 'List item 1.')
('', 'List item 2.')
```

</details>

pafonta:
Note: This issue depends on #447. Indeed, the processing of tags is impacted by the decision taken on #447.

pafonta:
### `<table-wrap>`

This is a very complex element to parse if not with a specific goal in mind

For example, one would need to handle column spanning ([@colspan](https://jats.nlm.nih.gov/archiving/tag-library/1.2/attribute/colspan.html)): _how many column widths this cell will span_.

If the parsing is done naively, the output would have the following format (example, not exhaustive):

```
caption
column headers
row headers
cell values
```

[...]

PR #497, opened 2021-11-08
Improve JATS Parsing

### LINKED ISSUE #496: Fix parsing of <list> for JATS XML
## 🐛 Bug description

Currently, the elements of lists are duplicated by the parser `PMCXMLParser`.

⚠️  This bug could also happen for `<def-list>` (see below) or, actually, any other tag.

There are two possible types of lists: [`<list>`](https://jats.nlm.nih.gov/archiving/tag-library/1.2/element/list.html) and [`<def-list>`](https://jats.nlm.nih.gov/archiving/tag-library/1.2/element/def-list.html).

## To reproduce
Example XML of an article with a list in a paragraph:

```xml
<article>
    <body>
        <p>
            <list list-type="bullet">
                <list-item>
                    <p>List item 1.</p>
                </list-item>
                <list-item>
                    <p>List item 2.</p>
                </list-item>
            </list>
        </p>
    </body>
</article>
```

Parsing of such article file (article.xml):

```python
from bluesearch.database.article import JATSXMLParser

parser = JATSXMLParser("article.xml")
for paragraph in parser.paragraphs:
  print(paragraph)
```

Duplicated outputs:

```
('', 'List item 1.\n                List item 2.')
('', 'List item 1.')
('', 'List item 2.')
```

## Expected behavior

Output:

```
('', 'List item 1.')
('', 'List item 2.')
```

EmilieDel:
Fixes #496, #425 and #429.

## Description

The PR improves JATS Parsing:
- Do not `yield` empty paragraphs
- Replace `\xa0` and `\u2009` by spaces (check  abstract of `/raid/projects/bbs/pmc/comm_use/Trauma_Surg_Acute_Care_Open/PMC6045722.nxml` for concrete illustration)
- Replace double-slash by single-slash when possible. However, the structure is always changing so it is not possible to put single slash everywhere (for example the number of sections varies `./body/sec/sec/.../sec/p/`)
- For paragraphs into paragraphs, concatenate all children paragraphs into one big one. (still to be discussed)

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/article.py
+++ b/src/bluesearch/database/article.py
@@ -22,11 +22,12 @@ import string
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Generator, Iterable, Optional, Sequence, Tuple
+from typing import Generator, Iterable, Optional, Sequence, Tuple, overload
 from xml.etree.ElementTree import Element  # nosec
 
 from defusedxml import ElementTree
 from mashumaro import DataClassJSONMixin
+from typing_extensions import Literal
 
 from bluesearch.database.identifiers import generate_uid
 
@@ -184,16 +185,9 @@ class JATSXMLParser(ArticleParser):
         str
             The paragraphs of the article abstract.
         """
-        abstract_pars = self.content.findall("./front/article-meta/abstract//p")
-        children_p = []
-        for paragraph in abstract_pars:
-            # Check if paragraph is not child of another paragraph
-            children_p += paragraph.findall(".//p")
-            if paragraph not in children_p:
-                text = self._element_to_str(paragraph)
-                # Check if text is not empty
-                if text:
-                    yield text
+        abstract = self.content.find("./front/article-meta/abstract")
+        if abstract:
+            yield from self.parse_section(abstract, title=False)
 
     @property
     def paragraphs(self) -> Generator[tuple[str, str], None, None]:
@@ -208,30 +202,10 @@ class JATSXMLParser(ArticleParser):
         text : str
             The paragraph content.
         """
-        # Abstract are removed because already parsed in abstract property.
-        abstract_p = self.content.findall("./front/article-meta/abstract//p")
-        # Caption are removed because already parsed later in the current method
-        # with the section name "Caption".
-        caption_p = self.content.findall("./body//caption/p")
-        # Acknowledgements are removed because not useful.
-        acknowledg_p = self.content.findall("./back/ack//p")
-
-        exclude_list = abstract_p + caption_p + acknowledg_p
-
         # Paragraphs of text body
-        children_p = []
-        section_dirs = self.get_paragraphs_sections_mapping()
-        for paragraph in self.content.findall(".//p"):
-            # Check that paragraph is not in the exclude list and
-            # that paragraph is not child of another paragraph.
-            if paragraph not in exclude_list and paragraph not in children_p:
-                children_p += paragraph.findall(".//p")
-                text = self._element_to_str(paragraph)
-                section_title = ""
-                if paragraph in section_dirs:
-                    section_title = section_dirs[paragraph]
-                if text:
-                    yield section_title, text
+        body = self.content.find("./body")
+        if body:
+            yield from self.parse_section(body, title=True)
 
         # Figure captions
         figs = self.content.findall("./body//fig")
@@ -246,8 +220,8 @@ class JATSXMLParser(ArticleParser):
         # Table captions
         tables = self.content.findall("./body//table-wrap")
         for table in tables:
-            caption_elements = table.findall(".//caption/p") or table.findall(
-                ".//caption/title"
+            caption_elements = table.findall("./caption/p") or table.findall(
+                "./caption/title"
             )
             if caption_elements is None:
                 continue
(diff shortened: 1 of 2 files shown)


src/bluesearch/database/article.py
```diff
@@ -184,9 +184,16 @@ def abstract(self) -> Generator[str, None, None]:
...
+        children_p = []
         for paragraph in abstract_pars:
-            yield self._element_to_str(paragraph)
+            # Check if paragraph is not child of another paragraph
+            children_p += paragraph.findall(".//p")
+            if paragraph not in children_p:
+                text = self._element_to_str(paragraph)
+                # Check if text is not empty
+                if text:
+                    yield text
```
pafonta, line 196: Ending with such logic was what I was talking about when we discussed https://github.com/BlueBrain/Search/issues/449#issuecomment-961230794. Especially, ending with `children_p` to keep a state is exactly the point I was rising.

It feels cumbersome to me to have such a mix of sequential, recursive, and stateful logic. That's why I was advocating for simplifying the method, especially by rethinking the recursivity done by `_element_to_str` and `_inner_text`.

A simplification is to process everything sequentially (`for tag in self.content.find("./front/article-meta/abstract/")`).

pafonta (approved):
The example code from #496 is now running as expected. So, I approve the PR.

I have made though a comment below about the concern I have already expressed.

src/bluesearch/database/article.py
EmilieDel: Sorry I was not aware of this discussion. I can try to experiment and simplify the process.

pafonta: If others are OK with the current implementation, no need to change something. I have actually already approved.

jankrepl (approved):
LGTM. Thank you!

src/bluesearch/database/article.py
```diff
@@ -202,44 +209,51 @@ def paragraphs(self) -> Generator[tuple[str, str], None, None]:
...
+            # that paragraph is not child of another paragraph.
+            if paragraph not in exclude_list and paragraph not in children_p:
+                children_p += paragraph.findall(".//p")
                 text = self._element_to_str(paragraph)
                 section_title = ""
                 if paragraph in section_dirs:
                     section_title = section_dirs[paragraph]
-                yield section_title, text
+                if text:
+                    yield section_title, text
```
Stannislav, line 234: Similarly to above, I'm still not 100% happy with the way we handle this. Like PA noted, keeping exclusion lists while doing global globbing looks rather hacky and I can't tell with certainty there won't be any edge cases that break this.

I might not have the full picture here, but by looking at one XML file I picked I see the following structure:
```xml
<body>
    <sec>
        <title>Section 1</title>
        <p>...</p>
        <p>...</p>
    </sec>
    <sec>
        <title>Section 1</title>
        <p>...</p>
        <p>...</p>
    </sec>
</body>
```
So what speaks against doing something like this:
```python
for sec in self.content.findall("./body"):
    sec_title = self._element_to_string(sec.find("title"))
    for par in sec.findall("p"):
        par_text = self._element_to_string(par)
        yield sec_title, par_text
```
plus adjusting `_element_to_string` so that it's a no-op for figures and tables because we want to extract them separately?

src/bluesearch/database/article.py
EmilieDel: The problem is that the paths to all the paragraphs are not consistent (i.e. we can several sections, subsections, ...). So it is not enough to search for `"./front/article-meta/abstract/p"`. Unfortunately, I am not sure we can avoid searching for `//p`, which leads to problem of duplications of paragraphs (exactly as the example code of the issue). 

If you have a better way to handle it, I am totally up to implement it.

src/bluesearch/database/article.py
```diff
@@ -150,7 +150,7 @@ def title(self) -> str:
         str
             The article title.
         """
-        titles = self.content.find(".//title-group/article-title")
+        titles = self.content.find("./front/article-meta/title-group/article-title")
```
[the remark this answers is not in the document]
EmilieDel: 👍

src/bluesearch/database/article.py
EmilieDel: I totally agree this comment. 

Maybe some comments: 
- Unfortunately, the structure is not always as straightforward as the one shown in your example: Paragraphs might not be contained in a section, there can have subsections, ... Moreover, from my experience, it is not possible to have a simple `XPath` catching all the paragraphs without any double slash (as illustrated with the previous example). 
- Just to clarify the purpose of the implementation: it seems sometimes paragraphs are children of paragraphs. To avoid any duplication of text, I decided to go for a solution parsing the parent paragraphs and avoid parsing again the children ones. 
- The exclude list logic was not introduced by this PR.

Stannislav: Do we have examples of paragarphs in `<body>` that are not inside a section? I'd be curious to see that.

Apart from that, I would say that it shouldn't matter if there are subsections or paragraphs within paragraphs, or any other nested structure as this should all be already succcessfully handled by `_element_to_str`.

My general feeling is that JATS being a fixed schema, its structure should be 100% predictable and there shouldn't be any stray elements in places where we wouldn't expect them to be. That's the reason why I keep thinking it should be possible to do the parsing without globbing over the whole XML file.

src/bluesearch/database/article.py
Stannislav: Like I said below, I think we shouldn't really care what kind of inner nesting the top-level `<p>` have, as it should be covered by `_element_to_str`.

For example in this snippet
```xml
<p>
Some text
<subsection>
	<p>More text</p>
	<p>Even more text [<ref>some ref</ref>]</p>
</subsection>
After subsection
</p>
```
Just applying `_element_to_str` to only the top `<p>` should give all the text and we don't care about the inner `<subsection>` and `<p>` tags.

src/bluesearch/database/article.py
```diff
@@ -354,7 +368,7 @@ def _inner_text(self, element: Element) -> str:
             text_parts.append(self._element_to_str(sub_element))
             # don't forget the text after the sub-element
             text_parts.append(html.unescape(sub_element.tail or ""))
-        return "".join(text_parts).strip()
+        return "".join(text_parts).replace("\xa0", " ").replace("\u2009", " ").strip()
```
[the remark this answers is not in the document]
EmilieDel: Done in 68d96bd.

src/bluesearch/database/article.py
EmilieDel: @Stannislav What do you think about the new implementation ?

src/bluesearch/database/article.py
EmilieDel: @pafonta @Stannislav What do you think about the new implementation ?

src/bluesearch/database/article.py
Stannislav: Nice, I love it :)) Thanks for implementing. I didn't test though, just read the code.

If you want to avoid the complexity with varying return types and the corresponding overloads one could just always return tuples `(title, text)`. Then for the abstracts one would probably do something like this:
```python
for _, text in self.parse_section(abstract):
    yield text
```
I think I'm fine either way. Up to you/others to decide :)

Stannislav: Another small trick would be to avoid the checks `if abstract` and `if text` be letting `parse_section` accept `Element | None` and then check for `None` inside `parse_section`:
```python
def parse_section(element: Element | None):
    if element is None:
        return
```

src/bluesearch/database/article.py
pafonta: The implementation is much cleaner and easy to follow in my opinion!

⚠️  However, there is a regression regarding fixing #496. Indeed, instead of giving
```
('', 'List item 1.')
('', 'List item 2.')
```
now the code is gving
```
('', 'List item 1.\n                List item 2.')
```

EmilieDel: Do you think it is a problem solving this issue this way ? I have the feeling trying to go to the children paragraphs is going to make things more complicated and we risk to miss the part contained in the parent one. Moreover, I don't think it is a problem to have a bigger paragraphs than several small ones as we mainly we work with sentences. However, I am up to discuss this if you think it is better to keep children paragraphs :)

src/bluesearch/database/article.py
EmilieDel: Thanks for the suggestions!
✅ Done the first one in d71e62b.

src/bluesearch/database/article.py
pafonta: > Do you think it is a problem solving this issue this way ?

Not at the moment.

> However, I am up to discuss this if you think it is better to keep children paragraphs :)

That's OK for now.

EmilieDel: Perfect! Can I resolve this conversation ?

PR #506, opened 2021-11-22
Enable bulk PDF to XML conversion in CLI

### LINKED ISSUE #483: Implement a CLI command for bulk PDF to XML conversion
This is a follow-up to #482 

How to implement the bulk conversion of PDFs?

**Option 1**
Using the same command.
- Pros:
  - fewer commands, less clutter.
  - the `parse` and the `add` commands work this way too
- Cons:
  - the parameters get overloaded, e.g `input_pdf_file` can be a file or a directory
  - this leads to lots of if-elses in the implementation
  - how to deal with the output file name? for one PDF it should be specified, for bulk conversion automatic generation

**Option 2**
Using a new command, e.g. `convert-pdf-bulk`
- Pros:
  - will probably be a cleaner implementation
  - less nesting and if/else blocks, more linear
- Cons:
  - Diverges from how `add` and `parse` work
  - A new command, maybe some duplication in the implementation

_Originally posted by @Stannislav in https://github.com/BlueBrain/Search/issues/482#issuecomment-951730082_

Stannislav:
Fixes #483 .

## Description
So far the `bbs_database convert-pdf` command supported the converion of a single PDF file. Rewrite and extend this command to support processing many PDF files.

## How to test?
(will provide later)

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/src/bluesearch/entrypoint/database/convert_pdf.py
+++ b/src/bluesearch/entrypoint/database/convert_pdf.py
@@ -255,11 +255,11 @@ def _convert_pdf_file(
     output_path
         The output directory for the XML file.
     """
-    logger.info("Reading %s", input_path.resolve().as_uri())
+    logger.info(f"Reading {input_path.resolve().as_uri()}")
     with input_path.open("rb") as fh_pdf:
         pdf_content = fh_pdf.read()
 
-    logger.info("Converting %s to XML", input_path.resolve().as_uri())
+    logger.info(f"Converting {input_path.resolve().as_uri()} to XML")
     xml_content = grobid_pdf_to_tei_xml(pdf_content, grobid_host, grobid_port)
 
     with output_path.open("w") as fh_xml:
--- a/src/bluesearch/entrypoint/database/convert_pdf.py
+++ b/src/bluesearch/entrypoint/database/convert_pdf.py
@@ -141,6 +141,8 @@ def run(
         The path to the input PDF file or a directory with PDF files.
     output_dir
         The output directory for the XML files.
+    num_workers
+        The number of parallel workers.
     force
         If true overwrite the output file if it already exists.
 
--- a/src/bluesearch/entrypoint/database/convert_pdf.py
+++ b/src/bluesearch/entrypoint/database/convert_pdf.py
@@ -124,6 +124,19 @@ def run(
     int
         The exit code of the command
     """
+    logger.info("Initialising")
+    from bluesearch.database.pdf import grobid_is_alive, grobid_pdf_to_tei_xml
+
+    # Check the GROBID server
+    if not grobid_is_alive(grobid_host, grobid_port):
+        logger.error(
+            "The GROBID server is not alive; please check your connection "
+            "and server configuration"
+        )
+        return 1
+    else:
+        logger.info("The GROBID server is alive")
+
     # Check if the input file exists
     if not input_pdf_path.exists():
         logger.error(
@@ -146,8 +159,6 @@ def run(
 
     # Convert the PDF to XML
     logger.info("Converting PDF to XML")
-    from bluesearch.database.pdf import grobid_pdf_to_tei_xml
-
     xml_content = grobid_pdf_to_tei_xml(pdf_content, grobid_host, grobid_port)
 
     # Write the XML file
--- a/src/bluesearch/entrypoint/database/convert_pdf.py
+++ b/src/bluesearch/entrypoint/database/convert_pdf.py
@@ -18,12 +18,15 @@
 from __future__ import annotations
 
 import argparse
+import asyncio
 import logging
 import pathlib
 import textwrap
 from typing import Iterable
 
-from bluesearch.database.pdf import grobid_is_alive, grobid_pdf_to_tei_xml
+import aiohttp
+
+from bluesearch.database.pdf import grobid_is_alive, grobid_pdf_to_tei_xml, grobid_pdf_to_tei_xml_aio
 
 logger = logging.getLogger(__name__)
 
@@ -154,8 +157,13 @@ def run(
         input_paths = _keep_pdfs_only(input_path.iterdir())
 
     # Convert
-    for input_path in input_paths:
-        _convert_pdf_file(grobid_host, grobid_port, input_path, output_dir, force)
+    path_map = prepare_files(input_paths, output_dir, force)
+    if len(path_map) > 0:
+        output_dir.mkdir(exist_ok=True)
+        logger.info(f"Starting asynchronous PDF conversion: {len(path_map)} files")
+        asyncio.run(_convert_and_save_all(path_map, grobid_host, grobid_port))
+    else:
+        logger.info("No files to convert")
 
     return 0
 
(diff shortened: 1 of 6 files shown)


FrancescoCasalegno:
Could you also please have a look into trying to implement a parallel (have a look at `asyncio` and/or `threading`) loop of requests?

src/bluesearch/entrypoint/database/convert_pdf.py
```diff
@@ -63,22 +67,38 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
+        The output directory where the XML file(s) will be saved. If not
+        provided the output files will be placed in the same directory as
+        the input files.
+        """,
+    )
+    parser.add_argument(
+        "-w",
+        "--num-workers",
+        type=int,
+        default=5,
```
pafonta, line 90: Is there a rational behind the number 5?

src/bluesearch/entrypoint/database/convert_pdf.py
```diff
@@ -63,22 +67,38 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         help="The port of the GROBID server.",
     )
     parser.add_argument(
-        "input_pdf_path",
+        "input_path",
         type=pathlib.Path,
-        metavar="INPUT-PDF-PATH",
-        help="The path of the input PDF file.",
+        metavar="INPUT-PATH",
+        help="The path to a single PDF file or a directory with many PDF files.",
```
pafonta, line 73: Maybe this should be clarified that the 'globing' is recursive (line 163).

src/bluesearch/entrypoint/database/convert_pdf.py
```diff
@@ -124,38 +145,124 @@ def run(
...
-    logger.info("Writing the XML file to disk")
-    with output_xml_path.open("w") as fh_xml:
-        n_bytes = fh_xml.write(xml_content)
-    logger.info("Wrote %d bytes to %s", n_bytes, output_xml_path.resolve().as_uri())
+    if output_dir is not None:
+        output_dir.mkdir(exist_ok=True)
+
+    # Convert
+    def do_work(pdf_path):
+        xml_path = path_map[pdf_path]
```
pafonta, line 177: That might not be always thread-safe ([doc](https://docs.python.org/3/glossary.html#term-global-interpreter-lock)).

src/bluesearch/entrypoint/database/convert_pdf.py
```diff
@@ -124,38 +145,124 @@ def run(
...
+        try:
+            _convert_pdf_file(grobid_host, grobid_port, pdf_path, xml_path)
+        except Exception as exc:
+            logger.exception(
+                f"An error happened when processing {pdf_path.resolve().as_uri()}: "
+                f"{exc}"
+            )
+            # lists seem to be thread-safe in Python
+            failed_paths.append(pdf_path)
```
pafonta, line 186: That might not be always thread-safe ([doc](https://docs.python.org/3/glossary.html#term-global-interpreter-lock)).

src/bluesearch/entrypoint/database/convert_pdf.py
```diff
@@ -124,38 +145,124 @@ def run(
...
 
-    logger.info("PDF conversion done")
+    with ThreadPoolExecutor(max_workers=num_workers) as executor:
+        executor.map(do_work, path_map)
```
pafonta, line 189: Adding a value for `timeout` ([doc](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map)) will prevent the code to be stuck if the GROBID server is stuck.

src/bluesearch/entrypoint/database/convert_pdf.py
Stannislav: Good catch, absolutely not actually. We just put a random number in during dev and never changed it. I'm not sure what's a good default though. Maybe 1?

src/bluesearch/entrypoint/database/convert_pdf.py
Stannislav: Good catch, chaned in fc3a7c0

Stannislav: Good catch! Changed in fc3a7c0

src/bluesearch/entrypoint/database/convert_pdf.py
Stannislav: Good idea, I added a 60 second timeout in ac9d712, this should be generous enough for the PDF conversion.

Stannislav: However, we're not handling the `TimeoutError` that will be thrown in this case, but maybe that's exactly what we want? If the GROBID server is stuck then it's probably better to just crash?

src/bluesearch/entrypoint/database/convert_pdf.py
Stannislav: Yes, we did think about the thread-safety when writing this.

The article you linked is just the definition of the GIL, but I'm not seeing anything about thread-safety of lists/dicts there, or did I oversee something? Wouldn't the existence of GIL actually not make the opposite more likely?

Anyway, when we tried to research the same question we came across this [SO thread](https://stackoverflow.com/questions/6319207/are-lists-thread-safe). There, people seem to imply that the two operations we do (reading a value from a dict and appending a value to a list) are thread-safe. What do you think?

src/bluesearch/entrypoint/database/convert_pdf.py
Stannislav: see above

src/bluesearch/entrypoint/database/convert_pdf.py
pafonta: Great!

src/bluesearch/entrypoint/database/convert_pdf.py
pafonta: Both the documentation I have linked and the SO thread say that the thread-safety is only true on CPython. So, if someone use another implementation, for performances for example, this code will no more be thread-safe.

src/bluesearch/entrypoint/database/convert_pdf.py
pafonta: Great!

Yes, I guess so.

src/bluesearch/entrypoint/database/convert_pdf.py
FrancescoCasalegno: I pushed a couple of new commits so that there should no longer be any thread safety concern.
Does the code now look fine after 5725821?

src/bluesearch/entrypoint/database/convert_pdf.py
FrancescoCasalegno: Same as above, does the code now look fine after 5725821?

Issue #520, opened 2021-12-06
Implement `bbs_database topics-extract --source=arxiv`


FrancescoCasalegno:
### Scope
- In #513 we have implemented a command to download papers form `arXiv`, together with the metadata file `arxiv-metadata-oai.json`.
- The current Issue is about implementing the `bbs_database topic-extract` command to write topic infos in the format specified in https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160. 
- If the output `jsonl` file already exists, we should append lines to it, not re-write it from scratch.

### Notes
- The structure of the metadata file and of the downloaded papers is discussed in https://github.com/BlueBrain/Search/issues/473#issuecomment-953049870
- In #513 we have already tried to avoid downloading, e.g., quantum physics papers located in `arxiv/quant-ph/`. Indeed we only downloaded `arxiv/arxiv` and `arxiv/q-bio`. But still, starting from `2007-04`, `arXiv` has placed papers on any topic inside `arxiv/arxiv`, so we have downloaded lots of papers related to a variety of topics (see again https://github.com/BlueBrain/Search/issues/473#issuecomment-953049870).
- The `arxiv-metadata-oai.json` file contains all information about the topic of each article, so the path to this file should be provided to the command. However, we should be careful with this file:
   - We will have downloaded various versions of the same paper, but we want to anyway consider only the latest version (https://github.com/BlueBrain/Search/issues/473#issuecomment-958931384). So in the output `jsonl` file we should consider writing info only for a `path` corresponding to the latest version.
   - Some of the papers mentioned in the `arxiv-metadata-oai.json` may not be present among the ones we downloaded. This is because sometimes papers are removed from the `arXiv` database upon request of the authors. So before writing info to our output `topics.jsonl` about a given paper, we should check that the paper exists on our machine.
  - So we should also pass to the command the path to the `arXiv` papers we have downloaded.

FrancescoCasalegno:
<table>
    <tr>
        <th>Approach</th>
        <th>Pros 😊</th>
        <th>Cons 😞</th>
        <th>Runtime 🕑</th>
    </tr>
    <tr>
        <td><a href="https://arxiv.org/help/api/user-manual">arXiv public API</a></td>
        <td>
            <ul>
                <li>Can retrieve metadata for article(s) with given <code>arXiv</code> id(s).</li>
            </ul>
        </td>
        <td>
            <ul>
[...]

PR #524, opened 2021-12-07
Implement `bbs_database download` for arXiv

### LINKED ISSUE #513: Implement `bbs_database download --source=arxiv`
- In https://github.com/BlueBrain/Search/issues/473 we have seen how we can download articles from `arXiv/q-bio`.
- We now need to implement a command line to support `bbs_database download --source=arxiv`.
- `arXiv` publishes regular updates to their db on a daily bases, so it is possible to only download papers uploaded to `arXiv` on certain dates.
- The command should support a `--from-month=YYYY-MM` option to let users download all and only those articles that were uploaded to `arXiv` after a give month of a given year.
- Note that `arXiv` contains a lot of articles covering lots of topics and — as we saw [here](https://github.com/BlueBrain/Search/issues/473#issuecomment-953049870) — papers belonging to `q-bio` represent a tiny fraction of those. We can dramatically reduce the number of papers downloaded by fetching only papers from `arxiv/q-bio/` and `arxiv/arxiv/`.
The `arxiv/arxiv/` will also contain a lot of articles covering other topics that we don't care about, but we can fix that during the filtering phase. But even just the fact of skipping the various `arxiv/astro-ph` and `arxiv/alg-geom` from our download will save us some work.
- Note that we also need to download `metadata-v5/arxiv-metadata-oai.json` so that then during filtering we will have access to information about the topics of each article.

jankrepl:
Fixes #513 

## Description
This implements the `bbs_database download arxiv` command.

There are several open question to discuss:
- The downloads contain all published versions of a given article (`XXXv1.pdf`, `XXXv2.pdf`, ...)
  - what do we want to do with them?
  - when do we want to do it?
- Is it possible that a new article version is published later and added to the same folder? How do we check that? (if we want to do that at all)
- The current implementation shows an error if the starting date is earlier than April 2007, this is when the new arxiv IDs were introduced. It's easy to implement downloads from before April 2007, but:
  - They are already pre-sorted into categories, e.g. `q-bio`, while the new ones aren't
  - So do we only want to download `q-bio` in that case?
  - If yes, do we mix the `q-bio` download from before April 2007 with the more recent ones that come from across all categories?
  - If we do mix, will this affect the subsequent topic filtering?
- At this stage no metadata processing and filtering is done at all. As a consequence PDFs from all categories are downloaded
  - This is pretty wasteful both server and client side
  - Can we do better or are we happy with the current approach?
 
## How to test?
Try the dry run:
```sh
bbs_database download -n arxiv 2021-12 out > to-download.txt
less to-download.txt
```

Try the real download (you'll probably want to ctrl+C at some point, it's a lot):
```sh
bbs_database download -v arxiv 2021-12 out
```

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/download.py
+++ b/src/bluesearch/database/download.py
@@ -22,9 +22,9 @@ import re
 from datetime import datetime, timedelta
 from pathlib import Path
 
+import pandas as pd
 import requests
 from boto3.resources.base import ServiceResource
-from google.api_core.page_iterator import HTTPIterator
 from google.cloud.storage import Blob, Bucket
 
 logger = logging.getLogger(__name__)
@@ -219,7 +219,7 @@ def get_gcs_urls(
     bucket: Bucket,
     start_date: datetime,
     end_date: datetime | None = None,
-) -> dict[str, HTTPIterator]:
+) -> dict[str, list[Blob]]:
     """Get Google Cloud Storage urls.
 
     Parameters
@@ -234,7 +234,7 @@ def get_gcs_urls(
     Returns
     -------
     url_dict
-        Keys represent different months. Values are iterators holding blobs
+        Keys represent different months. Values are list of blobs
         corresponding to actual PDF files.
     """
     date_list = get_daterange_list(
@@ -250,7 +250,28 @@ def get_gcs_urls(
     for yearmonth in yearmonth_list:
         iterator = client.list_blobs(bucket, prefix=f"arxiv/arxiv/pdf/{yearmonth}")
 
-        url_dict[yearmonth] = iterator
+        # If more than one version is found, we only keep the last one
+        df = pd.DataFrame(
+            (
+                (
+                    el,
+                    el.name,
+                    el.name.rsplit("v", 1)[0],
+                    int(el.name.rsplit("v", 1)[1].split(".")[0]),
+                )
+                for el in iterator
+            ),
+            columns=["blob", "fullname", "article", "version"],
+        )
+
+        df_latest = df[["article", "version"]].groupby("article", as_index=False).max()
+        fullname_latest = df_latest["name"] = (
+            df_latest["article"] + "v" + df_latest["version"].astype(str) + ".pdf"
+        )
+
+        url_dict[yearmonth] = df.loc[
+            df["fullname"].isin(fullname_latest), "blob"
+        ].to_list()
 
     return url_dict
 
--- a/src/bluesearch/database/download.py
+++ b/src/bluesearch/database/download.py
@@ -306,3 +306,34 @@ def download_s3_articles(
             output_path = parent_folder / url.split("/")[-1]
             logger.info(f"Downloading {url}")
             bucket.download_file(url, str(output_path), {"RequestPayer": "requester"})
+
+
+def download_gcs_blob(blob: Blob, out_dir: Path, *, flatten: bool = False) -> None:
+    """Download a Google Cloud Storage blob.
+
+    Parameters
+    ----------
+    blob
+        The blob to download.
+    out_dir
+        The output directory.
+    flatten
+        If false (default) then the directory structure encoded in the blob
+        will be recreated, otherwise the downloaded file will be placed
+        directly into the output directory. For example, if the blob name
+        is "my_files/subdir/file.bin" and flatten is true then the file
+        will be downloaded to "<output_dir>/file.bin", otherwise it will be
+        placed into "<output_dir>/my_files/subdir/file.bin".
+    """
+    path = Path(blob.name)
+    if flatten:
+        path = out_dir / path.name
+    else:
+        path = out_dir / path
+    logger.debug(f"Downloading {blob.name} to {path}")
+    path.parent.mkdir(exist_ok=True, parents=True)
+    pdf_content = blob.download_as_bytes()
+    # Not using download_to_file because an empty file is still created even
+    # if the blob does not exist.
+    with path.open("wb") as fh:
+        fh.write(pdf_content)
--- a/src/bluesearch/database/download.py
+++ b/src/bluesearch/database/download.py
@@ -24,6 +24,8 @@ from pathlib import Path
 
 import requests
 from boto3.resources.base import ServiceResource
+from google.api_core.page_iterator import HTTPIterator
+from google.cloud.storage import Blob, Bucket
 
 logger = logging.getLogger(__name__)
 
@@ -209,6 +211,46 @@ def get_s3_urls(
         )
 
         url_dict[month_year] = [obj.key for obj in objects if obj.key.endswith(".meca")]
+
+    return url_dict
+
+
+def get_gcs_urls(
+    bucket: Bucket,
+    start_date: datetime,
+    end_date: datetime | None = None,
+) -> dict[str, HTTPIterator]:
+    """Get Google Cloud Storage urls.
+
+    Parameters
+    ----------
+    bucket
+        GCS bucket.
+    start_date
+        Starting date to download the incremental files (inclusive).
+    end_date
+        Ending date. If None, today is considered as the ending date (inclusive).
+
+    Returns
+    -------
+    url_dict
+        Keys represent different months. Values are iterators holding blobs
+        corresponding to actual PDF files.
+    """
+    date_list = get_daterange_list(
+        start_date=start_date,
+        end_date=end_date,
+        delta="month",
+    )
+    yearmonth_list = [date.strftime("%y%m") for date in date_list]
+
+    client = bucket.client
+
+    url_dict = {}
+    for yearmonth in yearmonth_list:
+        iterator = client.list_blobs(bucket, prefix=f"arxiv/arxiv/pdf/{yearmonth}")
+
+        url_dict[yearmonth] = iterator
 
     return url_dict
 
--- a/src/bluesearch/entrypoint/database/download.py
+++ b/src/bluesearch/entrypoint/database/download.py
@@ -212,7 +212,7 @@ def run(source: str, from_month: datetime, output_dir: Path, dry_run: bool) -> i
                 )
                 job_names[future] = blob.name
                 n_blobs += 1
-                n_bytes += blob.size
+                n_bytes += blob.size or 0
                 if n_blobs % 100 == 0:
                     progress_info(n_blobs, n_bytes)
             progress_info(n_blobs, n_bytes)
--- a/src/bluesearch/entrypoint/database/download.py
+++ b/src/bluesearch/entrypoint/database/download.py
@@ -177,7 +177,7 @@ def run(source: str, from_month: datetime, output_dir: Path, dry_run: bool) -> i
             logger.error(
                 "The papers from before April 2007 follow a different format "
                 "and can't be downloaded. Please contact the developers if you "
-                "need them. To process please re-run the command with a "
+                "need them. To proceed please re-run the command with a "
                 "different starting month."
             )
             return 1
--- a/src/bluesearch/entrypoint/database/download.py
+++ b/src/bluesearch/entrypoint/database/download.py
@@ -18,7 +18,9 @@
 import argparse
 import getpass
 import logging
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
+from itertools import chain
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


src/bluesearch/database/download.py
```diff
@@ -264,3 +306,34 @@ def download_s3_articles(
...
+    """
+    path = Path(blob.name)
+    if flatten:
+        path = out_dir / path.name
+    else:
+        path = out_dir / path
+    logger.debug(f"Downloading {blob.name} to {path}")
+    path.parent.mkdir(exist_ok=True, parents=True)
+    pdf_content = blob.download_as_bytes()
+    # Not using download_to_file because an empty file is still created even
```
jankrepl, line 336: Did not know about this. However, how would we ever get hold of an empty file? Would this problem ever show up in our pipeline?

src/bluesearch/entrypoint/database/download.py
```diff
@@ -163,7 +165,68 @@ def run(source: str, from_month: datetime, output_dir: Path, dry_run: bool) -> i
...
+
+        client = Client.create_anonymous_client()
+        bucket = client.bucket("arxiv-dataset")
+        if from_month < datetime(2007, 4, 1):
+            logger.error(
+                "The papers from before April 2007 follow a different format "
+                "and can't be downloaded. Please contact the developers if you "
+                "need them. To proceed please re-run the command with a "
+                "different starting month."
+            )
```
jankrepl, line 182: This is nice. However, I feel like we should also do the same thing with other sources since basically all of them changed their "storing logic" at one point in the past. Might be worth creating a separate PR for this.

src/bluesearch/entrypoint/database/download.py
```diff
@@ -163,7 +165,68 @@ def run(source: str, from_month: datetime, output_dir: Path, dry_run: bool) -> i
...
+                    download_gcs_blob,
+                    blob,
+                    output_dir,
+                    flatten=True,
+                )
+                job_names[future] = blob.name
+                n_blobs += 1
+                n_bytes += blob.size or 0
+                if n_blobs % 100 == 0:
+                    progress_info(n_blobs, n_bytes)
```
jankrepl, line 217: I guess it is nice, however, again, we don't have any multi-threading logic for the other sources and the fact that we start introducing it now is not very consistent.

I guess we leave it here and have a new PR where we generalize it to all sources (and that way one can pass `--n-jobs` via the CLI

src/bluesearch/database/download.py
Stannislav: Don't know if this can happen in our pipeline, maybe if a blob is deleted server-side between our listing and downloading it?

I saw it in my experiments where I inserted fake blobs that don't exist in the blob sequence and saw that empty files are created for them anyway.

src/bluesearch/entrypoint/database/download.py
Stannislav: Yeah, that's also one of the points I put into "to discuss" in the description. I think the problem here is not the downloading itself, which is easy, but the fact that we'd download `q-bio` only (or another category) since that's how they used to sort them. I wasn't sure how that would interect with our filtering.

src/bluesearch/entrypoint/database/download.py
Stannislav: It's really a massive speedup to use threads, I tried without before and sequential downloads are slower by a lot. So I think without this it would be way less useable.

I might be less fussed than others with the idea that everything has to be harmonized. But I'm not against following up on this and trying to implement a similar logic for other source types.

src/bluesearch/database/download.py
jankrepl: Fair enough!

src/bluesearch/entrypoint/database/download.py
jankrepl: I created an issue here #526. I am resolving this conversation for now:)

src/bluesearch/database/download.py
```diff
@@ -213,6 +215,46 @@ def get_s3_urls(
...
+    end_date
+        Ending date. If None, today is considered as the ending date (inclusive).
+
+    Returns
+    -------
+    url_dict
+        Keys represent different months. Values are iterators holding blobs
+        corresponding to actual PDF files.
+    """
+    date_list = get_daterange_list(
```
FrancescoCasalegno, line 240: As you wrote in the PR,  this `get_gcs_urls()` will fail if `start_date < datetime.datetime(2007, 04, 01)`.

So, what about adding a small check for that and, if the `start_date` is not ok then raise a `NotImplementedError` or `ValueError`?

FrancescoCasalegno (approved):
LGTM, thanks for this @jankrepl !

~Just suggested a check for the `startdate`, otherwise~ feel free to merge.

Edit: Nevermind the strikethrough-ed text, I noticed that this event is covered already at line 176 of `src/bluesearch/entrypoint/database/download.py`.

src/bluesearch/database/download.py
```diff
@@ -264,3 +306,34 @@ def download_s3_articles(
...
+        The blob to download.
+    out_dir
+        The output directory.
+    flatten
+        If false (default) then the directory structure encoded in the blob
+        will be recreated, otherwise the downloaded file will be placed
+        directly into the output directory. For example, if the blob name
+        is "my_files/subdir/file.bin" and flatten is true then the file
+        will be downloaded to "<output_dir>/file.bin", otherwise it will be
+        placed into "<output_dir>/my_files/subdir/file.bin".
```
EmilieDel, line 326: I am not really aware of the structure of `arxiv`. Are we sure we won't have any conflict in the naming if we allow flattening ? Also, is flattening not going to be an issue to be able to retrace back where the file comes from ? I am mainly thinking out loud here, but keeping a copy of the source as is might be simplify a comparison between the source and our local copy and knows what is downloaded and what is not.

EmilieDel (approved):
Without looking too much into the code, I would say I have similar concerns as @jankrepl, especially because we start treating differences sources with different approaches. 
PMC also changed (very recently) the structure of the files. If the user puts a date before the change, we currently simply start from the changing date. I am not saying one approach is better than the other but I definitely think it is better to keep consistency in terms of behaviour to avoid confusion in the future.

src/bluesearch/database/download.py
Stannislav: We can totally keep the folder structure if this helps us in the future.

I reckoned the file names are already unique by themselves and just based on that went for the simpler flat directory structure. But maybe keeping the directory structure is better?

Stannislav:
> Without looking too much into the code, I would say I have similar concerns as @jankrepl, especially because we start treating differences sources with different approaches.
> PMC also changed (very recently) the structure of the files. If the user puts a date before the change, we currently simply start from the changing date. I am not saying one approach is better than the other but I definitely think it is better to keep consistency in terms of behaviour to avoid confusion in the future.

Are you referring to the fact that downloads with date < 2007-04-01 are not going to work? If yes then it's a legit point. The way it works now (warning + return 1) is more of a placeholder since I couldn't make a decision on how this should work. Ideally we should follow up on this.

EmilieDel:
> Are you referring to the fact that downloads with date < 2007-04-01 are not going to work? If yes then it's a legit point. The way it works now (warning + return 1) is more of a placeholder since I couldn't make a decision on how this should work. Ideally we should follow up on this.

Yes! I see :) I think it could be nice to exchange ideas on that point. We had a similar problem for PMC and decided for a solution without being really convinced about it. Decide together on one solution and harmonize could be beneficial :)

src/bluesearch/database/download.py
EmilieDel: If we are sure the file names are unique, then it shouldn't be hard to trace back the file in the source and make sure something was downloaded or not. So we can keep it that way!

FrancescoCasalegno:
@jankrepl  
Another thing — this PR only downloads the contents of `arxiv/arxiv/`, but no metadata file.

Such a metadata file is actually needed for #520, and in particular we need an up-to-date version of this metadata file.

Questions:
1. Does it make more sense to download this metadata file during `bbs_database download` or during `bbs_database topics-extract`? (I was leaning more towards the former, but actually the second can also be fine)
2. You mentioned that the metadata file on GCS (`gs://arxiv-dataset/metadata-v5/arxiv-metadata-oai.json`) is not up-to-date, so how would do you suggest to do? Maybe use the code [here](https://github.com/mattbierbaum/arxiv-public-datasets/blob/a91018a0c0647ceea6fb6b2426519cebd5f0effc/arxiv_public_data/oai_metadata.py#L179) from `arxiv-public-dataset` project?

Stannislav:
> @jankrepl Another thing — this PR only downloads the contents of `arxiv/arxiv/`, but no metadata file.
> 
> Such a metadata file is actually needed for #520, and in particular we need an up-to-date version of this metadata file.
> 
> Questions:
> 
>     1. Does it make more sense to download this metadata file during `bbs_database download` or during `bbs_database topics-extract`? (I was leaning more towards the former, but actually the second can also be fine)
> 
>     2. You mentioned that the metadata file on GCS (`gs://arxiv-dataset/metadata-v5/arxiv-metadata-oai.json`) is not up-to-date, so how would do you suggest to do? Maybe use the code [here](https://github.com/mattbierbaum/arxiv-public-datasets/blob/a91018a0c0647ceea6fb6b2426519cebd5f0effc/arxiv_public_data/oai_metadata.py#L179) from `arxiv-public-dataset` project?

re metadata.

* `gs://arxiv-dataset/metadata-v5/arxiv-metadata-oai.json` I checked that using the code snippets in #513, the function `get_meta_records` prints the latest date found in the JSON file, and if I remember correctly it was something around April this year or something.
* A more up-to-date version could be downloaded directly from Kaggle. Strangely, for me, it has small file size than the one from GS (3GB vs 4.2GB), but I didn't investigate the cause. Maybe some data is missing there? But I did check that the latest entry is from Dec 2021. 
* The same metadata can be queried directly from the arxiv servers, maybe that's an alternative for incremental downloads?
  * Docs: https://arxiv.org/help/oa/index
  * Note in the docs they mention the server may send a `503 Retry-After` status, we'd need to handle it.
  * Otherwise the interface isn't too crazy via `sickle`, see `test_sickle` in #513 (does sickle already handle `503 Retry-After`?)
  * There are some caveats when it comes to timestamps - as discussed before. (equal to the last revision, but sometimes also arbitrary, see examples in the docstring in `test_sickle` in #513 )
  * Maybe one can also query by the arxiv ID?
  *

FrancescoCasalegno:
### Follow-ups (discussed 

- [ ] harmonize behavior for other sources: what happens if `--from` date is too far away in the past? (we should `return 1` and log an error)
- [ ] arxiv id to be used to unique id (since pubmed does not exist for arxiv papers)
- [ ] filter blobs before download, in order to keep only latest version

FrancescoCasalegno:
> I guess one can do `fake_client.list_blobs.side_effect = lambda bucket, prefix: fake_blobs_by_prefix[prefix]`
> 
> This way the `fake_clinet.list_blobs` is going to be a `Mock` instance and one could potentially use it for some asserts later (e.g. `assert_called_once()`).

Done in 146288c, so we now also test that `list_blobs()` is called exactly 3 times.

src/bluesearch/database/download.py
```diff
@@ -213,6 +215,67 @@ def get_s3_urls(
...
+        )
+
+        df_latest = df[["article", "version"]].groupby("article", as_index=False).max()
+        fullname_latest = df_latest["name"] = (
+            df_latest["article"] + "v" + df_latest["version"].astype(str) + ".pdf"
+        )
+
+        url_dict[yearmonth] = df.loc[
+            df["fullname"].isin(fullname_latest), "blob"
+        ].to_list()
```
Stannislav, line 274: Looks fine to me. Just to note that the original `HTTPIterator` is lazy (I think), so evaluating it might lead to a slowdown. But maybe there's no other way.

Not a must, but here's an alternative implementation without pandas:
```python
    import re

    pattern = re.compile(r"(.*)v(\d+).pdf")
    for yearmonth in yearmonth_list:
        blobs = {}
        versions = {}
        for blob in client.list_blobs(bucket, prefix=f"arxiv/arxiv/pdf/{yearmonth}"):
            # Assuming pattern will always match
            name, version = pattern.fullmatch(blob.name).groups()
            if name not in versions or versions[name] < int(version):
                blobs[name] = blob
                versions[name] = int(version)
        url_dict[yearmonth] = list(blobs.values())
```

tests/unit/database/test_download.py
```diff
@@ -180,13 +180,35 @@ def fake_download_file(Key, Filename, ExtraArgs):
...
     start_date = datetime(2021, 10, 1)
     end_date = datetime(2021, 12, 1)
-    blobs_by_month = get_gcs_urls(bucket, start_date, end_date)
+    blobs_by_month = get_gcs_urls(fake_bucket, start_date, end_date)
+
     assert set(blobs_by_month) == {"2110", "2111", "2112"}
-    assert client.list_blobs.call_count == 3
+    assert blobs_by_month["2110"] == fake_blobs_by_prefix["arxiv/arxiv/pdf/2110"]
+    assert blobs_by_month["2111"] == fake_blobs_by_prefix["arxiv/arxiv/pdf/2111"][-1:]
+    assert blobs_by_month["2112"] == fake_blobs_by_prefix["arxiv/arxiv/pdf/2112"][:-1]
```
Stannislav, line 214: Comparing lists means we care about the ordering. Do we care about the ordering?

src/bluesearch/database/download.py
FrancescoCasalegno: Unrolling the `HTTPIterator` certainly isn't ideal, but I also cannot see another way to do that... anyway it runs in just a few seconds, so hopefully it's not too much of a big deal.

I went with the `pandas` solution as I hoped in some speedup, but if you prefer I can switch to the `dict` solution you propose!

tests/unit/database/test_download.py
FrancescoCasalegno: True, we don't care. Fixed in c24fa2a.

EmilieDel (approved):

PR #534, opened 2021-12-23
Extract `arXiv ID`

### LINKED ISSUE #507: Article UID can be None
Our `bbs_database parse` takes the article UID as the output filename: [relevant line in our code](https://github.com/BlueBrain/Search/blob/8e0202d08993d961e73b39229e83c5aa1f1ee001/src/bluesearch/entrypoint/database/parse.py)

However, there are cases where the article UID can be `None`. Here's an example from `arXiv` where this is the case:
```python
import requests

from bluesearch.database.article import Article, TEIXMLParser
from bluesearch.database.pdf import grobid_pdf_to_tei_xml


def main():
    pdf_url = "https://arxiv.org/pdf/1603.01360.pdf"
    grobid_host = "dgx1.bbp.epfl.ch"
    grobid_port = 8070

    response = requests.get(pdf_url)
    tei_xml = grobid_pdf_to_tei_xml(response.content, grobid_host, grobid_port)
    parser = TEIXMLParser(tei_xml)
    article = Article.parse(parser)
    print("Article UID:", article.uid)
    print("TEI IDs in article:", parser.tei_ids)


if __name__ == "__main__":
    main()
```
Output
```
Article UID: None
TEI IDs in article: {'MD5': '034E2A6D13AD37C00D813B0E4DBEB02D'}
```

Observations
* The output filename for this article would be `None.json`
* Currently we compute `article.uid` as `generate_uid((self.pubmed_id, self.pmc_id, self.doi))`
* In this concrete example we don't even have the arXiv ID in the XML (see TEI IDs above)
* It would be good to have a UID that is not `None`.

### LINKED ISSUE #531: Use `arXiv ID` to determine `unique ID` in the `article` table
- When generating the unique ID `uid` for the `article` table, we currently take into account `DOI`, `PMC ID`, and `PubMed ID`.
https://github.com/BlueBrain/Search/blob/7dae95ebd57973adba75efbce55342e295b223bd/src/bluesearch/database/article.py#L126
- But `arXiv` articles that are not on PubMed will have `None` for all these IDs, becuase the `DOI` field is present in `arXiv` papers only if there is a link that _"resolves (links) to another version of the article, in a journal for example"_ ([source](https://arxiv.org/help/prep#doi)). Indeed, _"arXiv users may also add DOIs to their papers during the submission process or add it to an already submitted paper by adding a journal reference"_ ([source](https://blog.arxiv.org/2019/05/16/orcid-doi/)).
    - _Example 1_: The paper "The Ising Model on a Dynamically Triangulated [...]" was published on [arXiv](https://arxiv.org/abs/hep-lat/0105002) and then on [Nuclear Physics B](https://www.sciencedirect.com/science/article/abs/pii/S0550321301004059). So on its arXiv page we can see its DOI: [10.1016/S0550-3213(01)00405-9](https://doi.org/10.1016/S0550-3213(01)00405-9).
    - _Example 2_: The paper "Very Deep Convolutional Networks [...]" was only ever published on [arXiv](https://arxiv.org/abs/1409.1556), so it has no DOI.
- Therefore, we should include the `arXiv` ID in the computation of our unique ID.
- Another point to consider is whether we want to include the version (e.g. `v1`, `v2`, ...) in our `arXiv` ID. However, notice that technically the version _is not_ part of the arXiv ID.

FrancescoCasalegno:
Fixes #531.
Partially fixes #507.

## Description

- This PR introduces the function `get_arxiv_id()` to infer the `arXiv ID` from the file path.
- This function gets called when calling `bbs_database parse` with `--input-type=tei-xml-arxiv`, which is basically the good old `TEIXMLParser` with a tiny change when `is_arxiv` is set to `True`.
- This `arXiv ID` is now also used when we compute and assign the article `uid`.

## How to test?

See the unit tests that were modified or added.

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


src/bluesearch/database/article.py
```diff
@@ -630,14 +680,23 @@ class TEIXMLParser(ArticleParser):
 
     Parameters
     ----------
-    xml_content
-        The content of a TEI XML file.
+    path
+        The path to a TEI XML file.
+    is_arxiv
```
jankrepl, line 685: I guess another approach would be to just try to extract the `arxiv_id` each time. And for non arxiv papers we would just catch the exception raised inside of `get_arxiv_id`.

This way there would be no change in the CLI and the user would not have to worry about it.

However, this approach is also fine. Your call.

jankrepl: How likely is it to have "false matches" (files that have the arxiv naming scheme, however, are not arxiv papers)?

jankrepl (commented):
Great job! Readable and minimal:) I left some comments in the code:)

src/bluesearch/database/article.py
```diff
@@ -102,6 +141,17 @@ def pmc_id(self) -> Optional[str]:
         """
         return None
 
+    @property
+    def arxiv_id(self) -> Optional[str]:
```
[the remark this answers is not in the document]
FrancescoCasalegno: To be honest I have never liked `Optional[Type]`, as I think it is a bit of a "misnaming" — especially when it's for an argument, it's easy to think of it as an "optional argument", while this is very not what it means at all.
So I would even be more radical and propose to replace _everywhere_ in our our code `Optional[Type]` with `Type | None`.

<details>
  <summary>(Side note)</summary>
I find `str | None` also somehow misleading, as `type(None)` is not `None` but `NoneType` and therefore I would expect `str | NoneType`. The docs of `mypy` say  [here](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type) that it's because it "is shorter and reads better", and [this issue](https://github.com/python/mypy/issues/11288) adds that "`NoneType` is a mangled up mess in the type realm and should never be used in a typing context".
</details>

TL;DR I used `Optional[str]` here just because I have seen it being used more often than `str | None` in our codebase. But I think `Optional` is a terrible naming choice, and if everyone here agrees  I can replace `Optional[T]` with `T | None` everywhere.

src/bluesearch/database/article.py
```diff
@@ -123,7 +173,7 @@ def uid(self) -> Optional[str]:
             If at least one identifier exists, unique id is created. Otherwise,
             the returned uid is None.
         """
-        return generate_uid((self.pubmed_id, self.pmc_id, self.doi))
+        return generate_uid((self.pubmed_id, self.pmc_id, self.arxiv_id, self.doi))
```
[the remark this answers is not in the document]
FrancescoCasalegno: > So this change will effectively "invalidate" all the existing UID's we have generated up until now 

Yes, and indeed I had to change the hashes in our unit tests, because the previous ones were now invalid.
As you say it's not a huge deal right now, but we should definitely fix the design of the `uid` computation before starting the actual population of the database, otherwise it will be a mess.

Is there any other `id` we could think of using? Maybe the `bioRxiv` and `medRxiv` ones?
This discussion is related to https://github.com/BlueBrain/Search/issues/507 (so maybe @Stannislav  any idea on other cases when the `uid` could be `None`?), we should try to make sure that the `uid` can always be computed.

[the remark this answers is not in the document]
FrancescoCasalegno: > Also, since the `arxiv_id` will hold the version it can lead to us having multiple versions of the same paper in our database. I guess that is what we want, right?

It's a design choice, I went for including the version to clarify any ambiguity. See also what they say in the `arXiv` documentation [here](https://arxiv.org/help/arxiv_identifier) (in particular, what happens if you don't specify the version):
> The identifier `arXiv:YYMM.numbervV` provides a complete and unique citation for an arXiv article. Without the version number (e.g. `arXiv:YYMM.number`), the identifier refers to the most recent version of the article.

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,44 @@
...
+
+    # Old format, up to 2007-03, needs to look at the whole path:
+    # - some_path/arxiv/<archive>/<format>/YYMM/YYMMNNNv<version>.<ext>
+    # Note: format may contain "-"
+    pattern = re.compile(r"arxiv/([\w-]+)/\w+/\d{4}/(\d{7}v\d+)\.\w+\Z")
+    match = re.search(pattern, "/".join(path.parts[-5:]))
+    if match:
+        return f"arxiv:{match.groups()[0]}/{match.groups()[1]}"
+
+    raise ValueError(f"Could not extract arXiv ID from file path {path}\n")
```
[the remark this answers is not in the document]
FrancescoCasalegno: Added to docstring in 3267a01fad0df788f8dc7cfd3f423081d30a0aa9.

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,44 @@
...
+    if match:
+        return f"arxiv:{match.groups()[0]}"
+
+    # Old format, up to 2007-03, needs to look at the whole path:
+    # - some_path/arxiv/<archive>/<format>/YYMM/YYMMNNNv<version>.<ext>
+    # Note: format may contain "-"
+    pattern = re.compile(r"arxiv/([\w-]+)/\w+/\d{4}/(\d{7}v\d+)\.\w+\Z")
+    match = re.search(pattern, "/".join(path.parts[-5:]))
+    if match:
+        return f"arxiv:{match.groups()[0]}/{match.groups()[1]}"
```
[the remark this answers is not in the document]
FrancescoCasalegno: Changed in 27e862af7cf7f824bda944a1ccd5a758c039e402.

src/bluesearch/database/article.py
FrancescoCasalegno: Personally I would find it a bit against the "principle of least astonishment" if a TEI parser assumed by default that the file is coming from a very specific source and therefore its file path contains particular information that mat be used.

This is why I set `is_arxiv` to `False` by default.

src/bluesearch/database/article.py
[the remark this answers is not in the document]
jankrepl: Yeh, I agree. I would suggest creating a separate PR/issue for the replacement in the entire codebase.

src/bluesearch/database/article.py
jankrepl: Sure:)

src/bluesearch/database/article.py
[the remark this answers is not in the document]
jankrepl: I agree that we should make sure that the UID computation includes all relevant identifiers before we actually start populating the database. Again, I guess this could be addressed in a separate issue/PR.

src/bluesearch/database/article.py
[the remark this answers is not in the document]
FrancescoCasalegno: Switched to `str | None` in 80c2053.

Opened https://github.com/BlueBrain/Search/pull/537 to use `X | Y` consistently in our codebase.

src/bluesearch/database/article.py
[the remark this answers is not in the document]
FrancescoCasalegno: See https://github.com/BlueBrain/Search/issues/507 for follow-up discussion on this.

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,50 @@
...
+    ----------
+    https://arxiv.org/help/arxiv_identifier
+    """
+    path = Path(path)
+
+    # New format, since 2007-04, only needs path stem:
+    # - since 2015-01 have format YYMM.NNNNN (i.e. 5 digits)
+    # - up to 2014-12 have format YYMM.NNNN (i.e. 4 digits)
+    pattern = re.compile(r"\A(\d{4}\.\d{4}\d?v\d+)\Z")
+    match = re.search(pattern, path.stem)
```
Stannislav, line 64: 1. I think for single use regexes the compiling might be redundant as it will be compiled in `re.search` anyway. So I'd suggest writing `re.search(r"\A(\d{4}\.\d{4}\d?v\d+)\Z", path.stem)`.
2. When compiling patterns one can also write `match = pattern.search(path.stem)`.
3. What do `\A` and `\Z` do exactly? The docs say "Matches only at the beginning/end of string", so are they equivalent to "^" and "$"? If we want to match the whole string maybe we should consider using `re.fullmatch(r"\d{4}\.\d{4}\d?v\d+", path.stem)`.

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,50 @@
...
+    if match:
+        return f"arxiv:{match.groups()[0]}"
```
Stannislav, line 66: One can also write `match.group(0)`.

If we use `match = re.fullmatch(...)` without groups to match the whole string, then we can also write `match.string` to get the whole input string that matched.

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,50 @@
...
+
+    # Old format, up to 2007-03, needs to look at the whole path:
+    # - some_path/arxiv/<archive>/<format>/YYMM/YYMMNNNv<version>.<ext>
+    # Note: format may contain "-"
+    pattern = re.compile(r"arxiv/([\w-]+)/\w+/\d{4}/(\d{7}v\d+)\.\w+\Z")
+    match = re.search(pattern, "/".join(path.parts[-5:]))
+    if match:
+        match_groups = match.groups()
+        return f"arxiv:{match_groups[0]}/{match_groups[1]}"
```
Stannislav, line 75: Here also one could write `f"arxiv:{match.group(0)}/{match_group(1)}"`

Or for more readability also
```python
cat, id_ = match.groups()
return f"arxiv:{cat}/{id_}"
```

src/bluesearch/database/article.py
```diff
@@ -32,6 +33,50 @@
...
+
+    raise ValueError(f"Could not extract arXiv ID from file path {path}\n")
```
Stannislav, line 77: Do we need thew newline at the end?

src/bluesearch/database/article.py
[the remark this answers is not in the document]
Stannislav: Regarding potential `None`s, I think according to #507 the sources we currently have are all covered.

As we discussed, we can think about a fall-back strategy for external/random texts that always works, like hashing the article title/abstract or something like that. I guess this is one of the potential places where this could be implemented. Something like
```python
@property
def uid(self) -> str:
    try:
        uid = generate_uid((self.pubmed_id, self.pmc_id, self.arxiv_id, self.doi))
    raise ValueError:
        uid = hash_article(self)
    return uid
```

src/bluesearch/database/article.py
```diff
@@ -630,14 +686,23 @@ class TEIXMLParser(ArticleParser):
...
-        self.content = ElementTree.fromstring(xml_content)
+    def __init__(self, path: str | Path, is_arxiv: Optional[bool] = False):
+        path = Path(path)
+        with path.open() as fp:
+            self.content = ElementTree.fromstring(fp.read())
         self.tei_namespace = {"tei": "http://www.tei-c.org/ns/1.0"}
         self._tei_ids: dict[str, str] | None = None
+        if is_arxiv:
+            arxiv_id = get_arxiv_id(path)
+            self._arxiv_id = arxiv_id
```
Stannislav, line 703: Why don't we write `self._arxiv_id = get_arxiv_id(path)`?

src/bluesearch/database/article.py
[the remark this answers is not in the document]
FrancescoCasalegno: @Stannislav  To avoid scope creep, I copied your comment to #507 and I suggest we keep discussing it there.

src/bluesearch/database/article.py
FrancescoCasalegno: In fact the whole block 
https://github.com/BlueBrain/Search/blob/4089e1756f9de6acbe26d2a7707a2a292659e80a/src/bluesearch/database/article.py#L701-L705
can be replaced by a one-line conditional expression.

Changed in 078ce32.

src/bluesearch/database/article.py
FrancescoCasalegno: Fixed in 2789b9e.

src/bluesearch/database/article.py
FrancescoCasalegno: Fixed in ec2ad5f.

src/bluesearch/database/article.py
FrancescoCasalegno: Fixed in ec2ad5f.

src/bluesearch/database/article.py
FrancescoCasalegno: 1. Yes, you only save the time of a lookup. I mainly used it for a style choice, to separate the pattern definition from the match search. From https://stackoverflow.com/a/452143:
> Python internally compiles and caches regexes whenever you use them anyway (including calls to re.match()), so you're really only changing when the regex gets compiled, and shouldn't be saving much time at all - only the time it takes to check the cache.

2. and 3. — Cool I didn't know about `pattern.search` and `fullmatch`! Changed in ec2ad5f.

PR #533, opened 2021-12-21
Implement `bbs_database topic-extract --source=pubmed`

EmilieDel:
Fixes #522.

## Checklist

- [X] This PR refers to an issue from the [issue tracker](https://github.com/BlueBrain/Search/issues).
  (if it is not the case, please create an issue first).
- [x] Unit tests added.
  (if needed)
- [x] Documentation and `whatsnew.rst` updated.
  (if needed)
- [x] `setup.py` and `requirements.txt` updated with new dependencies.
  (if needed)
- [x] Type annotations added.
  (if a function is added or modified)
- [x] All CI tests pass.

PR #535, opened 2022-01-03
Implement `topic-extract` for bioRxiv and medRxiv

### LINKED ISSUE #521: Implement `bbs_database topics-extract --source=medrxiv` and `--source=biorxiv`
### Scope
- In #512 we have implemented a command to download papers form `biorxiv` and `medrxiv`.
- The current Issue is about implementing the `bbs_database topic-extract` command to write topic infos in the format specified in https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160. 
- If the output `jsonl` file already exists, we should append lines to it, not re-write it from scratch.

### Notes
- `medRxiv` and `bioRxiv` organize their papers by Collections into "Subject Areas", see [here](https://www.medrxiv.org/) and [here](https://www.biorxiv.org/).
- Only one "Subject Area" can be set for each paper, as mentioned in their FAQs [here](https://www.merxiv.org/about/FAQ) and [here](https://www.biorxiv.org/about/FAQ).
- Once you open an article, the web interface of `medRxiv` and `bioRxiv` clearly shows the "Subject Area" of it. So a hacky way to get the "Subject Area" could be to scrape the page of an article like [this one](https://www.biorxiv.org/content/10.1101/2021.11.23.469739v2) to get the element in the `span` tag with `class="highwire-article-collection-term"`, which in this case would be `"Neuroscience"`.
- But ideally we don't want to do this manual, time-consuming (and potentially dangerous: we may be blocked?) scraping.
  So can we get info on the "Subject Area" somewhere else, e.g. in the paper itself or in a metadata file we can download (similar to `arXiv` with `arxiv-metadata-oai.json`, see https://github.com/BlueBrain/Search/issues/473#issuecomment-953049870)?

jankrepl:
Fixes #521 

## Description
* Takes raw `.meca` files (just zip files), finds the relevant `.xml` inside and extracts the "Subject Area" from it (as can be seen on the websites - https://www.medrxiv.org/ and https://www.biorxiv.org/)
* The internal behavior is identical for both bioRxiv and medRxiv, however, for consistency the CLI allows for two separate sources `biorxiv` and `medrxiv`.
* The actual regex pattern was created after inspecting manually multiple different XMLs and trying to find a stable pattern.  I did not stumble upon an article where it wouldn't work, however, I only tested it on like 50 medRxiv + 50 bioRxiv papers.

## How to test?
You can first download some bioRxiv/medRxiv. However, you you need AWS login credentials and you will be charged. Alternatively, you can just use the files we recently downloaded on GPFS.

```bash
bbs_database topic-extract  -v  -o -m '.*\.meca' medrxiv medrxiv/Current_Content/April_2021 res.jsonl     
```

## Comments
* Working on this PR made me realize that specifically for the medRxiv and bioRxiv sources our current pipeline has a gap. The `bbs_database download` downloads raw `.meca` (=zip) files, the `bbs_databse topic-extract` extracts the topic from the `.meca` file, however, the `bbs_database parse` seems to require an `.xml` file rather than the `.meca` file. Therefore, there needs to be a step of "unzipping" the `.meca` file and finding the unique `content/doi_kind_of.xml` inside of it. However, this should be straightforward and probably can be done purely via shell scripting. 

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


src/bluesearch/database/topic.py
```diff
@@ -295,3 +297,59 @@ def get_topics_for_pmc_article(
...
+            if x.startswith("content/") and x.endswith(".xml")
+        ]
+
+        if len(xml_files) != 1:
+            raise ValueError(
+                "There needs to be exactly one .xml file inside of content/"
+            )
+
+        xml_file = xml_files[0]
+        pattern_topic = """<subj-group subj-group-type="hwp-journal-coll">\r\n<subject>(.*)</subject>"""  # noqa
```
jankrepl, line 338: Ugly, too many characters. Feel free to suggest alternatives

src/bluesearch/database/topic.py
```diff
@@ -295,3 +297,63 @@ def get_topics_for_pmc_article(
...
+            match_journal = re.search(pattern_journal, text)
+
+            if match_topic is None:
+                raise ValueError("No topic found")
+
+            if match_journal is None:
+                raise ValueError("No journal found")
+
+            topic = match_topic.groups()[0]
+            journal = match_journal.groups()[0]
```
Stannislav, line 357: Instead of `groups()[0]` one can write `group(0)`.

src/bluesearch/database/topic.py
Stannislav: * Maybe using single quotes? `r'<subj-group subj-group-type="hwp-journal-coll">\r\n<subject>(.*)</subject>'`
* I'd use raw strings for patterns
* Is it necessary to write "\r\n"?
* Maybe it's safer to use `defusedxml` to parse this? Or is it too slow?

src/bluesearch/database/topic.py
```diff
@@ -295,3 +297,63 @@ def get_topics_for_pmc_article(
...
+            text = f.read().decode("utf-8")
+            match_topic = re.search(pattern_topic, text)
+            match_journal = re.search(pattern_journal, text)
+
+            if match_topic is None:
+                raise ValueError("No topic found")
+
+            if match_journal is None:
+                raise ValueError("No journal found")
+
```
EmilieDel, line 355: Just out of curiosity, did you test this on real data ? Did you encounter the case when only one of the two variables is `None` ? Do you think we should still raise an error in that context ?

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -168,6 +171,30 @@ def run(
...
                 }
             )
+
+    elif source in {"biorxiv", "medrxiv"}:
+        for path in inputs:
+            logger.info(f"Processing {path}")
+            topic, journal = extract_info_from_zipfile(path)
+            all_results.append(
+                {
+                    "source": journal,
```
EmilieDel, line 181: I think the `source` identified for the other cases was to simply put the source itself (i.e. `biorxiv`, `medrxiv`, `pmc`, `pubmed`, `arxiv`).

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -168,6 +171,30 @@ def run(
...
+                    "path": str(path.resolve()),
+                    "topics": {
+                        "article": {
+                            "Subject Area": topic,
+                        },
```
EmilieDel, line 186: `extract_info_from_zipfile(path)` is extracting the Subject Area and the `journal title`. Do you know if there is an easy way to extract some topic information about the journal ? Do you think it could be something worth to investigate ?

EmilieDel (approved):
LGTM! 

Just out of curiosity, when looking at this, did you see some ways to extract this subject area before downloading the article ? It could be useful to avoid downloading unused articles.

src/bluesearch/database/topic.py
jankrepl: * Raw strings - great point - it seems like it is the best practise
* "\r\n" - yeh, it is necessary since all the XML are using this carriage return + new line separator
* `defusedxml` - I was tying to avoid "parsing" the entire XML since in our pipeline it is done in a completely different step. I am not really that familiar with `defusedxml` and I am not sure if it can act "lazily". The fact that `re.search` only looks for the first occurrence was the main reason why I went for it.  However, great point! IMO if there are some issues with this `re` approach or it becomes unreadable we should stop using it.

jankrepl: The single quotes and raw strings were done here: 1b99e9d78cd5470536db7057bd3a586b43536326

src/bluesearch/database/topic.py
jankrepl: It should be `group(1)` right (unit tests didn't like it)? https://docs.python.org/3/library/re.html#re.Match.group

Anyway, done in d7bddd0

jankrepl:
> LGTM!
> 
> When looking at this, did you see some ways to extract this subject area before downloading the article ? It could be useful to avoid downloading unused articles.

Yeh, that is a reasonable suggestion. However, AFAIK our pipeline always stars with raw download of "everything" (no matter the input source) and it is only then that we try to kick out some articles that do not have the right topic. However, I might be wrong. What do you think @EmilieDel ?

src/bluesearch/database/topic.py
jankrepl: Yeh, I tested it on real data and I actually never stumbled upon a paper that would have any of the above entries missing. However, maybe it will happen. Under the current setup, this would completely stop the execution of the entrypoint. Do we have some convention in place on what to put inside of the output `.jsonl` in case the topic extraction failed? And should I implement it?

src/bluesearch/entrypoint/database/topic_extract.py
jankrepl: That is a possibility. However, what is tricky is that no matter if the user provided biorxiv` or `medrix` the same code is executed (= the extraction logic is identical for both of them). And that is why I just wanted to make sure that we actually find the journal inside of the `xml` rather then just copying what the CLI `source` was. 

I hope that makes sense. Maybe this logic is stupid. What do you think?

src/bluesearch/entrypoint/database/topic_extract.py
jankrepl: Great point. I did not investigate it all. Do you think it would be worth investigating? Intuitively, I would guess that article level topic should be more specific than the journal level one.

src/bluesearch/database/topic.py
Stannislav: I see the point about no parsing the whole XML. Main motivation for using a parser would be to not have to worry about equivalent syntax like `\n` vs `\r\n` and `"stuff"` vs `'stuff'` etc.

src/bluesearch/database/topic.py
Stannislav: Ah good point!

src/bluesearch/entrypoint/database/topic_extract.py
Stannislav: Also see the implementation in #539 I guess it would make sense to be consistent on this one.

jankrepl: @Stannislav makes sense. I am about to review it actually. However, maybe we can do it in a separate PR once both this PR and the other one are merged.

src/bluesearch/database/topic.py
jankrepl: Yeh, let's keep this one open and let's discuss it during the planning.

src/bluesearch/entrypoint/database/topic_extract.py
Stannislav: Yeah, I also don't mind merging this PR first and then adjusting mine

src/bluesearch/database/topic.py
FrancescoCasalegno: ### Discussed on 2022-01-11

- Note that we are already doing an `XML` parsing for `topic-extract` when the source is `PubMed` or `PMC` (see #525 and #533)
- Yes, there is some duplication of the parsing work, it is true
- Try to benchmark the runtime of extracting the topic from a pretty long `XML` file using the two approaches (`regex` vs. `XML` parsing) — if the parsing doesn't take too long, then I think we can go with that w/o worrying too much.

src/bluesearch/entrypoint/database/topic_extract.py
EmilieDel: As discussed during sprint, let's keep it that way :)

EmilieDel:
Yes you are right. I was asking because in this case we are charged for the download. But I think it is reasonable to keep the way we do currently.

src/bluesearch/database/topic.py
EmilieDel: I think for pmc we put a `None` to the value when there was a problem related to requests and an empty list when nothing was found but the pipeline went through. However, I am not 100% sure anymore. We can maybe keep it this way for now and adapt it later if needed.

FrancescoCasalegno (commented):
Awesome work digging into the mysteries of these `.meca` zipped files!

Everything good from my point of view, the only thing is the `regex` vs `parse` thing we discussed yesterday: https://github.com/BlueBrain/Search/pull/535#discussion_r782257838

src/bluesearch/database/topic.py
```diff
@@ -297,6 +299,66 @@ def get_topics_for_pmc_article(
     return journal_topics
 
 
+def extract_info_from_zipfile(path: pathlib.Path) -> tuple[str, str]:
```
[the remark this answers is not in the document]
jankrepl: Makes sense. I just noticed that out naming is not consistent actually.

```
git checkout master
git grep 'def extract' -- src/bluesearch/database/

src/bluesearch/database/topic.py:def extract_pubmed_id_from_pmc_file(path: str | pathlib.Path) -> str | None:
src/bluesearch/database/topic.py:def extract_article_topics_for_pubmed_article(
src/bluesearch/database/topic.py:def extract_journal_topics_for_pubmed_article(
```
Sometimes we use "for" and sometimes we use "from". But maybe this can be handled in a separate PR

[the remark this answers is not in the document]
jankrepl: I used the name `extract_article_topics_from_medrix_article `and the fix is here: df022eb916b124776191d45df8f54bd1117e1731

src/bluesearch/database/topic.py
```diff
@@ -297,6 +299,66 @@ def get_topics_for_pmc_article(
...
+    """Extract topic of a medRxiv/bioRxiv article.
+
+    The `.meca` file should always have a fixed structure. Namely,
+    there is a folder `content` and inside of it there should be
+    a single `.xml` file containing the text and the metadata of the
+    article.
+
+    Parameters
+    ----------
+    path
```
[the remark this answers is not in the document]
jankrepl: Done  in`f870aab093fb4131c23acfacb83845f97ee58797`

jankrepl:
> Awesome work digging into the mysteries of these `.meca` zipped files!
> 
> Everything good from my point of view, the only thing is the `regex` vs `parse` thing we discussed yesterday: [#535 (comment)](https://github.com/BlueBrain/Search/pull/535#discussion_r782257838)

Regarding the benchmark:

Locally, I have downloaded ~100 XMLs (they lie inside of `biomed_topic/`)

```
bbs_database topic-extract  -v --recursive -o -m '.*\.meca' -n biorxiv biomed_topic/ res.jsonl | wc -l                                │         return topic, journal
     101          
```

The original regex logic is for example in this commit f870aab093fb4131c23acfacb83845f97ee58797

```
git checkout f870aab093fb4131c23acfacb83845f97ee58797
time bbs_database topic-extract   --recursive -o -m '.*\.meca'  biorxiv biomed_topic/ res.jsonl

real    0m0.669s
user    0m2.117s
sys     0m0.490s
```


Whereas the parsing logic is in this commit 04634604fceaf9cb22928490c6e44e172436bce5

```
git checkout 04634604fceaf9cb22928490c6e44e172436bce5
time bbs_database topic-extract   --recursive -o -m '.*\.meca'  biorxiv biomed_topic/ res.jsonl


real    0m0.912s
user    0m2.268s
sys     0m0.489s
```

src/bluesearch/database/topic.py
jankrepl: See https://github.com/BlueBrain/Search/pull/535#issuecomment-1014457144 for the benchmark

FrancescoCasalegno:
> The above results are of course fluctuating a little bit but the regex solution seems to be consistently a little bit faster than the parsing one. However, IMO it is completely negligible and we can definitely use parsing since imo it makes the code more readable.

@jankrepl awesome work, thanks for this extensive benchmarking!
As we discussed last week with the rest of the team, I would go with the parsing, which seems to be only marginally slower but with the benefit of increased readability.

jankrepl:
> > The above results are of course fluctuating a little bit but the regex solution seems to be consistently a little bit faster than the parsing one. However, IMO it is completely negligible and we can definitely use parsing since imo it makes the code more readable.
> 
> @jankrepl awesome work, thanks for this extensive benchmarking! As we discussed last week with the rest of the team, I would go with the parsing, which seems to be only marginally slower but with the benefit of increased readability.

Great! The parsing is contained in the latest commit. So unless nobody protests we can merge this PR.

PR #536, opened 2022-01-03
Ignore deprecation warning emitted by the docker package

jankrepl:
This should be a temporary fix for https://github.com/docker/docker-py/issues/2928.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -138,6 +138,7 @@ commands =
 testpaths = tests
 filterwarnings =
     error
+    ignore::DeprecationWarning:docker.*:
 addopts =
     --cov
     --cov-config=tox.ini
--- a/tox.ini
+++ b/tox.ini
@@ -136,6 +136,7 @@ commands =
 
 [pytest]
 testpaths = tests
+# reason for the ignore: https://github.com/docker/docker-py/issues/2928
 filterwarnings =
     error
     ignore::DeprecationWarning:docker.*:


FrancescoCasalegno:
For future reference — this fix is needed because `setuptools` has deprecated in https://github.com/pypa/setuptools/commit/1701579e0827317d8888c2254a17b5786b6b5246 the usage of `distutils` `Version` classes, which are still used in the `docker` Python package ([here](https://github.com/docker/docker-py/blob/a9748a8b702a3c75b46ba8c8d0490e4b8ec5ab04/docker/utils/utils.py#L8) and [here](https://github.com/docker/docker-py/blob/a9748a8b702a3c75b46ba8c8d0490e4b8ec5ab04/docker/transport/ssladapter.py#L7)).

FrancescoCasalegno (approved):
Awesome, thanks for fixing this.
Let's merge asap so our tests go back to running 😁

tox.ini
```diff
@@ -138,6 +138,7 @@ commands =
 testpaths = tests
 filterwarnings =
     error
+    ignore::DeprecationWarning:docker.*:
```
Stannislav, line 141: Nice. Could we maybe add a comment line with the link to the GitHub issue? So we can directly track if it's resolved.

jankrepl: Done in 2510a38

Stannislav (approved):

PR #537, opened 2022-01-03
Use modern type annotations (PEP 604 + PEP 585)

FrancescoCasalegno:
## Description

This PR rewrites type annotations in `src/` and `tests/` to follow the new style introduced in [PEP 604](https://www.python.org/dev/peps/pep-0604/) and [PEP 585](https://www.python.org/dev/peps/pep-0585/). Basically, this means using:
- the shorthand `Type1 | Type2` to represent type union — so we get rid of `typing.Union` and `typing.Optional`
- the built-ins collection types instead of the generic duplicates from `typing` — so we get rid of the depracated `typing.Dict`, `typing.List`, `typing.Set`, ...

## How to test?

The output of 
-  `git grep -rIn "typing.*" src/  | sort -u`
- ` git grep -rIn "typing.*" tests/ | sort -u`
should show no trace of `Optional`, `Union`, `Set`, `Dict`, `List`, `Tuple`.

**NOTE** — There's only one exception to this cleanup I did, consisting in the class `Article` which inherits from `DataClassJSONMixin`. See ~https://github.com/lidatong/dataclasses-json/issues/327~ issue https://github.com/Fatal1ty/mashumaro/issues/65 (edit: the first issue was created in the wrong repo).

## Checklist

- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/article.py
+++ b/src/bluesearch/database/article.py
@@ -23,7 +23,7 @@ import unicodedata
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Generator, Iterable, Sequence
+from typing import Generator, Iterable, Optional, Sequence, Tuple
 from xml.etree.ElementTree import Element  # nosec
 
 from defusedxml import ElementTree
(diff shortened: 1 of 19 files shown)


Stannislav (approved):
Nice, thanks :)

Fatal1ty:
@FrancescoCasalegno FYI, I just released mashumaro [2.11](https://github.com/Fatal1ty/mashumaro/releases/tag/v2.11) that supports PEP 604. Since the new union type syntax requires Python 3.10, it won't work in older versions.

PR #540, opened 2022-01-10
Implement `bbs_database topic-extract --source=arxiv`

### LINKED ISSUE #520: Implement `bbs_database topics-extract --source=arxiv`

FrancescoCasalegno:
Fixes #520.

## Description

## How to test?

Try to launch
```bash
bbs_database   topic-extract   arxiv   <path>   output.jsonl
``` 
with the following 2 different paths
1. old format: `<path-to-arxiv-downloads>/fulltext-dataset/arxiv/q-bio/pdf/0703/`
2. new format: `<path-to-arxiv-downloads>/fulltext-dataset/fulltext-dataset/arxiv/arxiv/1808/` — this should also log some `ERROR`s because of a couple of files with a weird extension

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/topic.py
+++ b/src/bluesearch/database/topic.py
@@ -329,11 +329,10 @@ def get_topics_for_arxiv_articles(
     [2] https://arxiv.org/category_taxonomy
     """
     # Get arXiv IDs of interest.
-    prefix = "arxiv:"  # Need to drop the initial "arxiv:" from the arXiv ID
     id_2_path: dict[str, pathlib.Path] = {}
     for p in arxiv_paths:
         try:
-            arxiv_id = get_arxiv_id(p)[len(prefix) :]
+            arxiv_id = get_arxiv_id(p, with_prefix=False)
             id_2_path[arxiv_id] = pathlib.Path(p)
         except ValueError as ve:
             logger.error(f"Failed ID extraction: {ve}")
--- a/src/bluesearch/database/topic.py
+++ b/src/bluesearch/database/topic.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import html
 import logging
 import pathlib
+import re
 from functools import lru_cache
 from typing import Iterable
 from xml.etree.ElementTree import Element  # nosec
@@ -27,7 +28,7 @@ from xml.etree.ElementTree import Element  # nosec
 import requests
 from defusedxml import ElementTree
 
-from bluesearch.database.article import JATSXMLParser
+from bluesearch.database.article import JATSXMLParser, get_arxiv_id
 
 logger = logging.getLogger(__name__)
 
@@ -295,3 +296,81 @@ def get_topics_for_pmc_article(
             journal_topics.append(descriptor["name"])
 
     return journal_topics
+
+
+# arXiv source
+def get_topics_for_arxiv_articles(
+    arxiv_paths: Iterable[pathlib.Path] | Iterable[str], batch_size: int = 400
+) -> dict[pathlib.Path, list[str]]:
+    """Extract journal topics of one or more arXiv article.
+
+    Parameters
+    ----------
+    arxiv_paths
+        Full paths to the arXiv articles to consider.
+
+    batch_size
+        Metadata are retrieved using the arXiv API [1] in batches of size
+        `batch_size`. Large batches values may create long request URLs that
+        cause the arXiv API to fail.
+
+    Returns
+    -------
+    article_topics : dict[pathlib.Path , list[str]]
+        Maps each of the paths to a list of corresponding arXiv article topics.
+        See [2] for an explanation of arXiv topics taxonomy.
+
+    References
+    ----------
+    [1] https://arxiv.org/help/api/user-manual
+    [2] https://arxiv.org/category_taxonomy
+    """
+    # Get arXiv IDs of interest.
+    prefix = "arxiv:"  # Need to drop the initial "arxiv:" from the arXiv ID
+    id_2_path: dict[str, pathlib.Path] = {}
+    for p in arxiv_paths:
+        try:
+            arxiv_id = get_arxiv_id(p)[len(prefix) :]
+            id_2_path[arxiv_id] = pathlib.Path(p)
+        except ValueError as ve:
+            logger.error(f"Failed ID extraction: {ve}")
+
+    # Retrieve metadata in batches
+    ns = {
+        "atom": "http://www.w3.org/2005/Atom",
+        "arxiv": "http://arxiv.org/schemas/atom",
+        "opensearch": "http://a9.com/-/spec/opensearch/1.1/",
+    }
+    base_url = "http://export.arxiv.org/api/query"
+    id_pattern = re.compile(r"\Ahttp://arxiv.org/abs/(.*)")
+    ids = list(id_2_path.keys())
+    article_topics = {}
+    for i_start in range(0, len(ids), batch_size):
+        # Get a slice of arXiv ids
+        i_end = i_start + batch_size
+        id_list = ids[i_start:i_end]
+
+        # Send request to arXiv API for our slice of arXiv ids
+        params = {
+            "id_list": ",".join(id_list),
+            "max_results": str(batch_size),
+        }
+        res = requests.get(base_url, params)
+        res.raise_for_status()
+
+        # Process response to retrieve arXiv ids and corresponding topics
+        etree = ElementTree.fromstring(res.text)
+        entries = etree.findall("./atom:entry", ns)
+        if len(entries) != len(id_list):
+            raise ValueError(
+                "Expected to find {len(id_list)} metadata, "
+                "but found {len(entries)}, for id_list = {id_list}"
+            )
+        for el in entries:
+            id_ = id_pattern.findall(el.find("./atom:id", ns).text)[0]
+            categories = [
+                categ.get("term") for categ in el.findall("atom:category", ns)
+            ]
+            article_topics[id_2_path[id_]] = categories
+
+    return article_topics
--- a/src/bluesearch/database/topic.py
+++ b/src/bluesearch/database/topic.py
@@ -320,6 +320,11 @@ def get_topics_for_arxiv_articles(
         Maps each of the paths to a list of corresponding arXiv article topics.
         See [2] for an explanation of arXiv topics taxonomy.
 
+    Raises
+    ------
+    ValueError
+        If the arXiv API does not return the correct number of metadata.
+
     References
     ----------
     [1] https://arxiv.org/help/api/user-manual
@@ -363,8 +368,8 @@ def get_topics_for_arxiv_articles(
         entries = etree.findall("./atom:entry", ns)
         if len(entries) != len(id_list):
             raise ValueError(
-                "Expected to find {len(id_list)} metadata, "
-                "but found {len(entries)}, for id_list = {id_list}"
+                f"Expected to find {len(id_list)} metadata, "
+                f"but found {len(entries)}, for id_list = {id_list}"
             )
         for el in entries:
             id_ = id_pattern.findall(el.find("./atom:id", ns).text)[0]
--- a/src/bluesearch/database/topic.py
+++ b/src/bluesearch/database/topic.py
@@ -369,11 +369,16 @@ def get_topics_for_arxiv_articles(
                 f"but found {len(entries)}, for id_list = {id_list}"
             )
         for el in entries:
-            id_ = id_pattern.fullmatch(el.find("./atom:id", ns).text).group(1)
-            categories = [
-                categ.get("term") for categ in el.findall("atom:category", ns)
-            ]
-            article_topics[id_2_path[id_]] = categories
+            atom_id = el.find("./atom:id", ns)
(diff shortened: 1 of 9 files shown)


src/bluesearch/database/topic.py
```diff
@@ -297,6 +297,88 @@ def get_topics_for_pmc_article(
     return journal_topics
 
 
+def get_topics_for_arxiv_articles(
+    arxiv_paths: Iterable[pathlib.Path] | Iterable[str], batch_size: int = 400
```
jankrepl, line 301: Just wondering about the `arxiv_paths` type...Would it not be possible to do `Iterable[pathlib.Path | str]`? If yes, not sure whether which one is the recommended one

jankrepl, line 301: Regarding the `batch_size` I would imagine the default of 400 is reasonable and does not lead to any problems, right? Since there is no way the CLI user could control it.

tests/data/arxiv_api_response.xml
```diff
@@ -0,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jankrepl, line 1: Seems like this contains real data. Another approach would be to just fill all the entries with "fake" text like we did here for example https://github.com/BlueBrain/Search/blob/master/tests/data/jats_article.xml. However, it is your call.

tests/data/arxiv_api_response.xml
```diff
@@ -0,0 +1,123 @@
...
+    <author>
+      <name>Thimo Rohlf</name>
+    </author>
+    <author>
+      <name>Stefan Bornholdt</name>
+    </author>
+    <arxiv:comment xmlns:arxiv="http://arxiv.org/schemas/atom">13 pages revTex, 17 figures (2 color)</arxiv:comment>
+    <link href="http://arxiv.org/abs/q-bio/0401024v1" rel="alternate" type="text/html"/>
+    <link title="pdf" href="http://arxiv.org/pdf/q-bio/0401024v1" rel="related" type="application/pdf"/>
+    <arxiv:primary_category xmlns:arxiv="http://arxiv.org/schemas/atom" term="q-bio.MN" scheme="http://arxiv.org/schemas/atom"/>
```
jankrepl, line 40: So the `arxiv:primary_category` seems to be always equal to the first `catogory`, right? And that is why we don't extract the `arxiv:primary_category` in the code...

Are we sure it is always going to be the case?

jankrepl (commented):
Great job! I left a few comments in the code:)

EmilieDel (approved):
Nice 🎉 Thank you for doing this PR!

src/bluesearch/database/topic.py
```diff
@@ -297,6 +297,88 @@ def get_topics_for_pmc_article(
...
+        # Process response to retrieve arXiv ids and corresponding topics
+        etree = ElementTree.fromstring(res.text)
+        entries = etree.findall("./atom:entry", ns)
+        if len(entries) != len(id_list):
+            raise ValueError(
+                f"Expected to find {len(id_list)} metadata, "
+                f"but found {len(entries)}, for id_list = {id_list}"
+            )
+        for el in entries:
+            id_ = id_pattern.findall(el.find("./atom:id", ns).text)[0]
```
Stannislav, line 373: I would write
```python
id_pattern = re.compile(r"http://arxiv.org/abs/(.*)")
id_ = id_pattern.fullmatch(el.find("./atom:id", ns).text).group(1)
```
`fullmatch` already implies from the beginning of the string, so `\A` is not necessary.

tests/unit/database/test_topic.py
```diff
@@ -384,6 +386,50 @@ def test_get_topics_for_pmc_article(test_data_path, monkeypatch):
...
+    with open(test_data_path / "arxiv_api_response.xml") as f:
+        body = f.read()
+    id_queries = [
+        "id_list=q-bio%2F0401024v1%2Cq-bio%2F0401014v1%2C1808.02949v2",
+        "id_list=q-bio%2F0401024v1%2Cq-bio%2F0401014v1%2C1808.02949v2%2C1808.02950v7",
+    ]
+    for id_query in id_queries:
+        responses.add(
+            responses.GET,
+            f"http://export.arxiv.org/api/query?{id_query}&max_results=400",
```
Stannislav, line 400: @jankrepl you looked into `responses` recently - wasn't there a new syntax for adding URL paramters?

jankrepl: The problem we had was related to the function `responses.add_callback`. AFAIK the `responses.add` function should be fine.

https://github.com/BlueBrain/Search/pull/546

Stannislav: I mean something like in the example here: https://github.com/getsentry/responses#query-parameters-as-a-dictionary Maybe there wasn't an explicit usecase for URL parameters in #546.

In the example I linked they do say "Note, do not use query parameters as part of the URL".

jankrepl: oops, sorry, I misunderstood.

Yeh, one should use on of the options below instead of making it part of the URL
* `matchers.query_param_matcher`
* `matchers.query_string_matcher`

src/bluesearch/database/topic.py
FrancescoCasalegno: As @Stannislav pointed out, they mean different things and the one you are suggesting covers more cases.
But luckily for us either type would work in this case, and I like your proposal because it's more compact.
Fixed in b063949.

FrancescoCasalegno: In my experiments `400` seemed indeed a good number that allowed to retrieve large batches without having an error response from the server:
- in principle, the `arXiv` API can retrieve up to `1,000` different items per query
- but if you concatenate too many `arXiv id`s values in the query, then the URL becomes too long and the `arXiv` server returns error.

---

> there is no way the CLI user could control it

This is true, but I couldn't really think of a clean way to implement it.
- Adding a `batch_size` parameter to `argparse` would only make this parameter applicable to the case of `arxiv` source, while for any other sources it would be ignored.
- We could have a sort of `**kwargs` capture of extra parameters passed by command line and try to unpack it when calling the various methods, but this idea of capturing arbitrary parameters with `argparse` seems to be a bad design practice as it makes things not very transparent (see e.g. comments [here](parameter)).

What would you suggest?

jankrepl: I think it is OK the way it is now.

One way of handling extra arguments would be https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args

However, as you mentioned I would avoid doing if possible to avoid extra complexity.

jankrepl: True, they mean something different:) Thanks!

src/bluesearch/database/topic.py
```diff
@@ -297,6 +297,88 @@ def get_topics_for_pmc_article(
...
+    ValueError
+        If the arXiv API does not return the correct number of metadata.
+
+    References
+    ----------
+    [1] https://arxiv.org/help/api/user-manual
+    [2] https://arxiv.org/category_taxonomy
+    """
+    # Get arXiv IDs of interest.
+    prefix = "arxiv:"  # Need to drop the initial "arxiv:" from the arXiv ID
```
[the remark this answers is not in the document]
FrancescoCasalegno: Added `with_prefix` parameter (with some unit tests) in 7104c44.

src/bluesearch/database/topic.py
FrancescoCasalegno: Changed in 3fcf20b.

tests/data/arxiv_api_response.xml
FrancescoCasalegno: Removed the titles, authors, abstract contents in 87f4494.

tests/data/arxiv_api_response.xml
FrancescoCasalegno: Yes, the `arxiv:primary_category` should always be also one of the `category` items, which is why it does not make sense to extract it a second time.


> seems to be always equal to the first `category`

I would not go as far as assuming anything on the order of the `<category>` elements, I think the primary category could also be not the first `<category>` element. But this should be irrelevant, unless we actually care about knowing which is the primary category.

jankrepl: I see:) Thank you!

tests/unit/database/test_topic.py
FrancescoCasalegno: Cool, I did not know about this new syntax!

Notice that we are already using `responses.add` with `URL` string arguments including query parameters.
So if we change it here let's do it everywhere.

FrancescoCasalegno: Also, notice that `responses` introduced these `matchers` in `0.14.0` (see [here](https://github.com/getsentry/responses/releases)) so we need to bump our requirements.

FrancescoCasalegno: Done in 72f1231.

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -174,6 +175,25 @@ def run(
...
+                    },
+                },
+                "metadata": {
+                    "created-date": datetime.datetime.now().strftime(
+                        "%Y-%m-%d %H:%M:%S"
+                    ),
+                    "bbs-version": bluesearch.version.__version__,
+                },
+            }
+            for path, article_topics in get_topics_for_arxiv_articles(inputs).items()
```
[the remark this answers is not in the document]
FrancescoCasalegno: The `ValueError` is currently not caught, as you indeed correctly observed. The reason for this is that when that exception is raised it means that there's some issue with the `arXiv` server. This cannot be fixed as it does not depend on the client and represents, imo, a case for raising a fatal error that interrupts the CLI command execution.

I agree that we should try to harmonize the behavior across sources in terms of how exceptions are handled!

src/bluesearch/database/topic.py
```diff
@@ -297,6 +297,88 @@ def get_topics_for_pmc_article(
...
+        res = requests.get(base_url, params)
+        res.raise_for_status()
+
+        # Process response to retrieve arXiv ids and corresponding topics
+        etree = ElementTree.fromstring(res.text)
+        entries = etree.findall("./atom:entry", ns)
+        if len(entries) != len(id_list):
+            raise ValueError(
+                f"Expected to find {len(id_list)} metadata, "
+                f"but found {len(entries)}, for id_list = {id_list}"
```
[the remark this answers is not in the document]
FrancescoCasalegno: This exception here is a just a sanity check before actually trying to find the metadata info about the different `id`s requests to the server. 

The thing is that this exception here is being raised without even looking into content of these `entries` and trying to extract the `arXiv id`s (see few lines after this line), so when we raise this exception we don't even know what are the returned `id`s, we just know that their count does not match our request.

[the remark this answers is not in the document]
EmilieDel: I see, thank you for the explanation! Sounds good for me!

src/bluesearch/entrypoint/database/topic_extract.py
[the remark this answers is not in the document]
EmilieDel: I see, I thought the error could also come from the data itself. If it is a server problem, I totally agree that the CLI should be interrupted!

src/bluesearch/database/topic.py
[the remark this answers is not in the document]
Stannislav: Awesome, cheers!

Stannislav (approved):
LGTM

Issue #518, opened 2021-11-30
Define output format for `bbs_database topic extract`


FrancescoCasalegno:
## Context
- The various literature archives we use to download papers (`PubMed`, `PMC`, `arXiv`, ...)  include a lot of papers that are not related to NeuroScience.
- It would be hard to manage a literature database that includes all sources, so we need to identify the topic of each article to then be able to filter out those that are not of interest.
- Notice that more advanced filtering (e.g. consider only papers in English, or published after a certain date, ...) can anyway be done afterwords, and possibly even at runtime. Here we are just trying to identify the topics of each article.

## Scope
- Before we implement a `bbs_database topic extract` command (dedicated Issues will be created), the present Issue requires to define the output format of such a command.
  - Should we use `CSV`, `JSONL`, or other?
  - How do we deal with the fact that a single file (typically this is the case of `PubMed`) may contain more than one article? 
  - If CSV, which columns do we want in our schema? E.g. `source, filename, article_n_in_file, topic_info`?
  - How do we handle the fact that the topic info depend on the source? E.g. for `PMC` we have a list of MeSH topics relative to the journal and a list of MeSH topics relative to the article, while for `arXiv` we have a list of arXiv topics relative to the article etc.

FrancescoCasalegno:
I think CSV may not be the most suitable to this use case, because:
- sets of topics are more naturally interpreted as lists and tabular format is not very list-friendly
- we have attributes that depend on the source, e.g. if an article is not from `arXiv` all topics lists will be empty
- the filed `article_n_in_file` is only relevant for `PubMed`, and in all other cases it will be a `0`


What about a JSONL like this?

<details>
  <summary>Example of an element of the JSONL representing a PubMed paper</summary>
  
  ```json
  {
    "filename": "/gpfs/.../proj115/downloads/pubmed/some_file.txt", 
    "article_n_in_file":  1,
    "source": "pubmed"
    "mesh_topics_article": 
      [
        "A08", 
        "C10.10", 
        "E04.525"
      ],
    "mesh_topics_journal":
      [
        "A08", 
        "C10"
      ]
  }
  ```
</details>


<details>
  <summary>Example of an element of the JSONL representing an arXiv paper</summary>
  
  ```json
  {
    "filename": "/gpfs/.../proj115/downloads/arXiv/some_file.pdf", 
    "source": "arxiv"
    "arxiv_topics": 
      [
        "cs.CV", 
        "math.DS"
      ]
  }
  ```
[...]

PR #542, opened 2022-01-10
Add the TopicInfo data structure

### LINKED ISSUE #539: Add a data structure for topic info
We're starting to copy-paste a nested dictionary with a fixed structure that represents the extracted journal/article topics, e.g.:
```python
{
    "source": "pubmed",
    "path": str(path.resolve()),
    "topics": {
        "journal": {
            "MeSH": journal_topics,
        },
        "article": {
            "MeSH": article_topics,
        },
    },
    "metadata": {
        "created-date": datetime.datetime.now().strftime(
            "%Y-%m-%d %H:%M:%S"
        ),
        "bbs-version": bluesearch.version.__version__,
        "element_in_file": i,
    },
}
```
It could be appropriate to introduce a data structure do deal with this instead.

For more context/details see this comment: https://github.com/BlueBrain/Search/pull/533#discussion_r778227276

Stannislav:
Fixes #539 .

## Description
This adds the `TopicInfo` data class that replaces the manual creation and copy-pasting of the structured dictionary that we have been using for this purpose so far.

## How to test?
* See new tests under `tests/unit/database/test_topic_info.py`
* Adjusted tests for old code pass

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -1,0 +1,116 @@
+# Blue Brain Search is a text mining toolbox focused on scientific use cases.
+#
+# Copyright (C) 2020  Blue Brain Project, EPFL.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+"""Implementation of the TopicInfo data structure."""
+from __future__ import annotations
+
+import copy
+import datetime
+import pathlib
+from dataclasses import dataclass, field
+from typing import Any
+
+import bluesearch
+from bluesearch.database.article import ArticleSource
+
+
+@dataclass
+class TopicInfo:
+    """The topic information extracted from a journal article.
+
+    For the spec see the following GitHub issue/comment:
+    https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160
+    """
+
+    source: ArticleSource
+    path: str | pathlib.Path
+    element_in_file: int | None = None
+    article_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    journal_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    metadata: dict = field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        """Run the post-initialization."""
+        self.creation_date = datetime.datetime.now()
+
+    @staticmethod
+    def add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
+        """Add topics to a mapping with collection of topics.
+
+        Parameters
+        ----------
+        mapping
+            A mapping of the form kind -> list-of-topics that shall be
+            updated in-place. For example ``{"MeSH": ["topic 1", "topic 2"]}``.
+        kind
+            The topic kind. Corresponds to a key in ``mapping``.
+        topics
+            The topics to add. Corresponds to a value in ``mapping``.
+        """
+        updated_topics = mapping.get(kind, []) + topics
+        mapping[kind] = sorted(set(updated_topics))
+
+    def add_article_topics(self, kind: str, topics: list[str]) -> None:
+        """Add article topics.
+
+        Parameters
+        ----------
+        kind
+            The topic kind. For example "MeSH" or "MAG".
+        topics
+            A list of the topics to add.
+        """
+        self.add_topics(self.article_topics, kind, topics)
+
+    def add_journal_topics(self, kind: str, topics: list[str]) -> None:
+        """Add journal topics.
+
+        Parameters
+        ----------
+        kind
+            The topic kind. For example "MeSH" or "MAG".
+        topics
+            A list of the topics to add.
+        """
+        self.add_topics(self.journal_topics, kind, topics)
+
+    def json(self) -> dict:
+        """Convert the contents of this class to a structured dictionary.
+
+        Returns
+        -------
+        dict
+            The structure dictionary with all topic information.
+        """
+        metadata: dict[str, Any] = {
+            **copy.deepcopy(self.metadata),
+            "created-date": self.creation_date.strftime("%Y-%m-%d %H:%M:%S"),
+            "bbs-version": bluesearch.__version__,
+        }
+        if self.element_in_file is not None:
+            metadata["element_in_file"] = self.element_in_file
+
+        json = {
+            "source": self.source.value,
+            "path": str(self.path),
+            "topics": {
+                "article": copy.deepcopy(self.article_topics),
+                "journal": copy.deepcopy(self.journal_topics),
+            },
+            "metadata": metadata,
+        }
+
+        return json
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -40,7 +40,6 @@ class TopicInfo:
     element_in_file: int | None = None
     article_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
     journal_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
-    metadata: dict = field(default_factory=dict)
 
     def __post_init__(self) -> None:
         """Run the post-initialization."""
@@ -96,7 +95,6 @@ class TopicInfo:
             The structure dictionary with all topic information.
         """
         metadata: dict[str, Any] = {
-            **copy.deepcopy(self.metadata),
             "created-date": self.creation_date.strftime("%Y-%m-%d %H:%M:%S"),
             "bbs-version": bluesearch.__version__,
         }
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -90,6 +90,10 @@ class TopicInfo:
     def json(self) -> dict:
         """Convert the contents of this class to a structured dictionary.
 
+        Apart from the source, path and topic entries a "metadata" top-level
+        key will be added containing a dictionary with entries "created-date"
+        and "bbs-version".
+
         Returns
         -------
         dict
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -44,6 +44,7 @@ class TopicInfo:
     def __post_init__(self) -> None:
         """Run the post-initialization."""
         self.creation_date = datetime.datetime.now()
+        self.path = pathlib.Path(self.path).resolve()
 
     @staticmethod
     def add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -47,7 +47,7 @@ class TopicInfo:
         self.path = pathlib.Path(self.path).resolve()
 
     @staticmethod
-    def add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
+    def _add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
         """Add topics to a mapping with collection of topics.
 
         Parameters
@@ -73,7 +73,7 @@ class TopicInfo:
         topics
             A list of the topics to add.
         """
-        self.add_topics(self.article_topics, kind, topics)
+        self._add_topics(self.article_topics, kind, topics)
 
     def add_journal_topics(self, kind: str, topics: list[str]) -> None:
         """Add journal topics.
@@ -85,7 +85,7 @@ class TopicInfo:
         topics
             A list of the topics to add.
         """
-        self.add_topics(self.journal_topics, kind, topics)
+        self._add_topics(self.journal_topics, kind, topics)
 
     def json(self) -> dict:
         """Convert the contents of this class to a structured dictionary.
--- a/src/bluesearch/database/topic_info.py
+++ b/src/bluesearch/database/topic_info.py
@@ -47,7 +47,9 @@ class TopicInfo:
         self.path = pathlib.Path(self.path).resolve()
 
     @staticmethod
-    def _add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
+    def _add_topics(
+        mapping: dict[str, list[str]], kind: str, topics: list[str]
+    ) -> None:
         """Add topics to a mapping with collection of topics.
 
         Parameters
--- a/src/bluesearch/entrypoint/database/topic_extract.py
+++ b/src/bluesearch/entrypoint/database/topic_extract.py
@@ -177,26 +177,11 @@ def run(
                 if journal_topics:
                     topic_info.add_journal_topics("MeSH", journal_topics)
                 all_results.append(topic_info.json())
-    elif source == "arxiv":
-        all_results = [
-            {
-                "source": "arxiv",
-                "path": str(path.resolve()),
-                "topics": {
-                    "article": {
-                        "arXiv": article_topics,
-                    },
-                },
-                "metadata": {
-                    "created-date": datetime.datetime.now().strftime(
-                        "%Y-%m-%d %H:%M:%S"
-                    ),
-                    "bbs-version": bluesearch.version.__version__,
-                },
-            }
-            for path, article_topics in get_topics_for_arxiv_articles(inputs).items()
-        ]
-
+    elif article_source == ArticleSource.ARXIV:
+        for path, article_topics in get_topics_for_arxiv_articles(inputs).items():
+            topic_info = TopicInfo(source=article_source, path=path)
+            topic_info.add_article_topics("arXiv", article_topics)
+            all_results.append(topic_info.json())
     elif source in {"biorxiv", "medrxiv"}:
         for path in inputs:
             logger.info(f"Processing {path}")
--- a/src/bluesearch/entrypoint/database/topic_extract.py
+++ b/src/bluesearch/entrypoint/database/topic_extract.py
@@ -121,11 +121,8 @@ def run(
     Parameter description and potential defaults are documented inside of the
     `get_parser` function.
     """
-    import datetime
-
     from defusedxml import ElementTree
 
-    import bluesearch
     from bluesearch.database.topic import (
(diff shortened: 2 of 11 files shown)


src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+            **copy.deepcopy(self.metadata),
+            "created-date": self.creation_date.strftime("%Y-%m-%d %H:%M:%S"),
+            "bbs-version": bluesearch.__version__,
+        }
+        if self.element_in_file is not None:
+            metadata["element_in_file"] = self.element_in_file
+
+        json = {
+            "source": self.source.value,
+            "path": str(self.path),
```
EmilieDel, line 108: Maybe we can keep the absolute path here ?

Update: I just saw that you are taking the absolute path when instantiating the TopicInfo into the entrypoint. Don't you think it could be worth to still take the absolute path here to be sure it is harmonized ?

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+        dict
+            The structure dictionary with all topic information.
+        """
+        metadata: dict[str, Any] = {
+            **copy.deepcopy(self.metadata),
+            "created-date": self.creation_date.strftime("%Y-%m-%d %H:%M:%S"),
+            "bbs-version": bluesearch.__version__,
+        }
+        if self.element_in_file is not None:
+            metadata["element_in_file"] = self.element_in_file
```
EmilieDel, line 104: Just for me to understand, what do we expect inside `self.metadata` ? 
Shouldn't `element_in_file` be part of the metadata  ?

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+
+        json = {
+            "source": self.source.value,
+            "path": str(self.path),
+            "topics": {
+                "article": copy.deepcopy(self.article_topics),
+                "journal": copy.deepcopy(self.journal_topics),
```
EmilieDel, line 111: 1. Just for me to understand, why do you think it is needed to do a `deepcopy` here ? 
2. Just to put into light, implementing the `topics` value this way means that we will have an `article` and `journal` field even when we do not search for one of them. Do you think it is a problem ?

EmilieDel (approved):
Really nice 🎉  It is definitely more clear and clean than copy-pasting the dictionary for every source, thanks for the changes!

src/bluesearch/database/topic_info.py
Stannislav: 1. It's to make sure you can't mutate your `TopicInfo` by changing the returned dictionary. Without deepcopy if your do `json = topic_info.json(); json["topic"]["article"]["x"] = 5` then this will also change the internal state of `topic_info`.
2. Yes, this is actually a slightly different behaviour to what we used to do. Do you know if we rely on the presence/abcense of the "article"/"journal" keys? I couldn't see it specified anywhere. I reconed it would be easier to deal with if it's always present and of the same type, so you don't have to constantly check for Nones.

src/bluesearch/database/topic_info.py
Stannislav: That's certainly possible. I left the responsibility for checking/resolving the path to the caller for two reasons
* We accept `pathlib.Path` and `str` types, so we'd have to convert back and forth
* If the path doesn't exist then resolving it could give unexpected results (from the [docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve): "[if] the path is resolved as far as possible and any remainder is appended without checking whether it exists")

But it's not an absolute no-go. What do the others think?

src/bluesearch/database/topic_info.py
Stannislav: Right now it's just possible to pass any additional metadata via `TopicInfo(..., metadata={"some key": "some value", ...})`, that's what will be in `self.metadata`. But if we don't want that, I guess we can get rid of that. I'm actually not so sure here about what we want exactly.

I'm not sure I understand the question about `element_in_file`? It's inserted into `metadata` if it's set, otherwise it will be absent. At least that's how it was specified in https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160.

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+    article_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    journal_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    metadata: dict = field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        """Run the post-initialization."""
+        self.creation_date = datetime.datetime.now()
+
+    @staticmethod
+    def add_topics(mapping: dict[str, list[str]], kind: str, topics: list[str]) -> None:
```
jankrepl, line 50: Maybe we can name it `_add_topics` to stress that it will only be used internally.

Completely subjective and feel free to ignore

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+        Returns
+        -------
+        dict
+            The structure dictionary with all topic information.
+        """
+        metadata: dict[str, Any] = {
+            **copy.deepcopy(self.metadata),
+            "created-date": self.creation_date.strftime("%Y-%m-%d %H:%M:%S"),
+            "bbs-version": bluesearch.__version__,
+        }
```
jankrepl, line 102: 1. Maybe we can document somewhere (class docstring?) that the `created-date` and `bbs-version` will be automatically added?
2. If the user is not aware of 1. they might accidentally provide `created-date` and `bbs-version` manually. Using the `{}` literal actually silently allows for duplicate keys. Maybe we can use `dict` instead to prevent this from happening (it raises an exception)

```python
In [13]: d_1 = {"a": 2}

In [14]: d_2 = {**d_1, "a": 2323}

In [15]: d_2
Out[15]: {'a': 2323}

In [16]: d_3 = dict(**d_1, a=2323)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-16-6dfbfab8a8f2> in <module>
----> 1 d_3 = dict(**d_1, a = 2323)

TypeError: type object got multiple values for keyword argument 'a'
```

But whatever, it is a detail.

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+    For the spec see the following GitHub issue/comment:
+    https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160
+    """
+
+    source: ArticleSource
+    path: str | pathlib.Path
+    element_in_file: int | None = None
+    article_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    journal_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
+    metadata: dict = field(default_factory=dict)
```
jankrepl, line 43: Might be nice to document these in the class docstring under `Attributes` (I guess that is the right section).

src/bluesearch/database/topic_info.py
jankrepl: IMO both are okay as long as we clearly document the behavior somewhere (e.g. class docstring).

jankrepl (commented):
Very reasonable! Thank you for the effort:)

FrancescoCasalegno:
### Discussed 2022-01-11

Let's wait for #535 and #540 to be merged into `master` before merging this PR.
Notice that the code will need to be changed to address the updates on the `master`.

src/bluesearch/database/topic_info.py
EmilieDel: Sorry, I am realising that my comment was not really clear and maybe not at the right place. I was wondering if `element_in_file` should be a separate attribute or if we should expect the user to give this info through the metadata dictionary (as it is really source specific and going to end in the metadata field anyway) but maybe it is better to keep it explicit.

src/bluesearch/database/topic_info.py
EmilieDel: 1. I see, makes sense. Thanks!
2. I don't think we currently directly rely on this. I thought it was going to be though to distinguish when we did not look for a topic and when looked for it but did not find any. But in the first case, the dictionary is going to be empty and in the second one, the dictionary is going to have a key with the value being an empty list. So maybe it is not really a problem. Moreover, we can still check the bbs code version.

src/bluesearch/database/topic_info.py
FrancescoCasalegno: I feel like Emilie that absolute paths are better, since we are then expecting to use this info when reading the file to the know which files should be loaded into the DB and which ones should not.

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -151,57 +146,38 @@ def run(
         print(*inputs, sep="\n")
         return 0
 
+    try:
+        article_source = ArticleSource(source)
+    except ValueError:
+        logger.error("Unknown article source: %s", source)
+        return 1
```
FrancescoCasalegno, line 153: I think it would be cleaner to handle at arg-parsing time this casting of the `str` passed on the command line into an `ArticleSource`.

This means the following small changes:
1. Specify the `type` in the `ArgumentParser`:
    ```python
    parser.add_argument(
        "source",
        type=ArticleSource,
        choices=(
            "arxiv",
            "biorxiv",
            "medrxiv",
            "pmc",
            "pubmed",
        ),
        choices=[member.value for member in ArticleSource.__members__.values()],
    ```
3. Change the signature of `run()` to take an `ArticleSource`:
    ```python
   def run(
      *,
      source: ArticleSource,
      input_path: Path,
      output_file: Path,
      match_filename: str | None,
      recursive: bool,
      overwrite: bool,
      dry_run: bool,
     ) -> int:
   ```
3. Remove the then unnecessary `try... except` block here.

src/bluesearch/database/topic_info.py
```diff
@@ -0,0 +1,116 @@
...
+        Parameters
+        ----------
+        kind
+            The topic kind. For example "MeSH" or "MAG".
+        topics
+            A list of the topics to add.
+        """
+        self.add_topics(self.journal_topics, kind, topics)
+
+    def json(self) -> dict:
```
FrancescoCasalegno, line 90: Wouldn't it make sense to leverage [`dataclasses.asdict`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) here?

FrancescoCasalegno (commented):
Thank you for your work @Stannislav Stan! 

It will definitely make our code simpler to maintain and understand!

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -45,13 +46,7 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
...
         "source",
         type=str,
-        choices=(
-            "arxiv",
-            "biorxiv",
-            "medrxiv",
-            "pmc",
-            "pubmed",
-        ),
+        choices=[member.value for member in ArticleSource.__members__.values()],
```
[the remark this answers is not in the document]
Stannislav: Cool, thanks! Commited in cba25d0

src/bluesearch/entrypoint/database/topic_extract.py
Stannislav: Yes, makes total sense. However, it seems that for some reason `argparse` does not play well with enums :( The use of enums is explicitly discouraged [in the docs](https://docs.python.org/3/library/argparse.html#choices). There's also a discussion about it [in the bug tracker](https://bugs.python.org/issue25061).

Long story short, this:
```python
parser.add_argument(
    "source",
    type=ArticleSource,
    choices=[member.name for member in ArticleSource],
```
doesn't work because argparse first applies the callable in the `type` argument and only then checks against the `choices`, which never matches because it compares `str` to `ArticleSource`.

This:
```python
parser.add_argument(
    "source",
    type=ArticleSource,
    choices=ArticleSource,
```
works, but the help message is misleading:
```
 bbs_database topic-extract --help
usage: bbs_database topic-extract [-h] [-v] [-m MATCH_FILENAME] [-R] [-o] [-n]
                                  {ArticleSource.ARXIV,ArticleSource.BIORXIV,ArticleSource.MEDRXIV,ArticleSource.PMC,ArticleSource.PUBMED,ArticleSource.UNKNOWN}
                                  input_path output_file
```
So one might have to stick with `str` type and casting to `ArticleSource` later. The `try-except` is probably not necessary if we use `choices`.

Stannislav: I removed the `try-except` in b1df1ae (+some cleaning up), but didn't didn't add `type=ArticleSource` otherwise.

src/bluesearch/database/topic_info.py
Stannislav: I'm not sure how we'd do it in our case since we return a dict with a different struture and also add additional values (creation date and bluesearch version) that are not members of the dataclass itself. Do you have a suggestion?

src/bluesearch/database/topic_info.py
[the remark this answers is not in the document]
Stannislav: I don't like this metadata either. I was following the spec here https://github.com/BlueBrain/Search/issues/518#issuecomment-985525160 where the idea of metadata was to contain any source-specific information.

I can actually remove it now because we don't have anything source-specific at the moment. Note though that the `metadata` will still be part of the dict returned by `TopicInfo.json()` and will contain a dictionay with `created-date`, `bbs-version`, and an optional `element_in_file` since this is what we put in there [on master](https://github.com/BlueBrain/Search/blob/d41611ad516e05b2fe28007659b4f33f3d6ce1a1/src/bluesearch/entrypoint/database/topic_extract.py) by hand.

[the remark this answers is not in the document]
Stannislav: Removed the `metadata` parameter in 827c79d

src/bluesearch/database/topic_info.py
Stannislav: Ah I see :) There's actually a discussion thread with @FrancescoCasalegno about the `metadata` parameter further below. For now I completely removed it in 827c79d. So I guess for now it has to stay as a separate paramter.

Stannislav: But feel free to chime in on the structure discussion below, I think we're still looking for a good solution :)

src/bluesearch/entrypoint/database/topic_extract.py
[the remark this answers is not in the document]
Stannislav: Good idea. Added in fb0d21f

src/bluesearch/database/topic_info.py
Stannislav: Sure, why not! Changed in 0ec90d4

src/bluesearch/database/topic_info.py
Stannislav: Actually, this `metadata` parameter got removed in 827c79d following an ongoing discussion with @FrancescoCasalegno below, so I guess this solves 2.

src/bluesearch/database/topic_info.py
Stannislav: Just tried and learned that `sphinx`/`autodoc`/`napoleon` don't play well with dataclasses. They're not clever enough to recognise them and document everything as class variables. If one includes a class docstring with `Parameters` and `Attributes` then sphinx generates duplicate entries and even issues warnings for `Attributes` because they are already implicitly included in the doc.

I haven't found a way to override this behaviour, there's a very ugly way of inline-documenting attributes/class variables:
```python
    source: ArticleSource
    path: str | pathlib.Path
    element_in_file: int | None = None
    #: A mapping of the form topic-kind -> list-of-topics that contains
    #: article topics. For example: ``{"MeSH": ["topic 1", "topic 2"]}``.
    #: Use ``add_article_topics`` method to populate it.
    article_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
    #: A mapping of the form topic-kind -> list-of-topics that contains
    #: journal topics. For example: ``{"MeSH": ["topic 1", "topic 2"]}``.
    #: Use ``add_journal_topics`` method to populate it.
    journal_topics: dict[str, list[str]] = field(init=False, default_factory=dict)
```
but to my eyes this looks extremely ugly.

I can think of two ways of going forward:
* Not documenting attributes of dataclasses at all
* Documenting paramters in the class docstring (but they will be duplicated as class variables because that's what sphinx thinks they are) and attributes inline like in the example above.

Any suggestions?

src/bluesearch/database/topic_info.py
Stannislav: Regarding 1. is the change in c9fc877  what you had in mind?

src/bluesearch/database/topic_info.py
Stannislav: Path resolution added here: 5145e04

src/bluesearch/database/topic_info.py
Stannislav: Actually, bad news, we can't make it private by adding an underscore:
```
src/bluesearch/database/topic_info.py:50:89: E501 line too long (89 > 88 characters)
```
🤣

src/bluesearch/database/topic_info.py
EmilieDel: Thank you! :)

src/bluesearch/entrypoint/database/topic_extract.py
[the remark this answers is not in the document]
jankrepl: Great! Thank you!

src/bluesearch/database/topic_info.py
jankrepl: 2. OK:)
1. Perfect, thanks!

src/bluesearch/database/topic_info.py
jankrepl: Pretty weird. I did not know about this issue. Yeh, maybe we don't have to write any docstrings. As you want:) Feel free to resolve without any action.

Stannislav: I'd leave it like it is for now. I think it relatively transparent right now. Maybe if we add more stuff to this class (cf. the metadata discussion) then it might be worth adding more documentation about it too.

src/bluesearch/entrypoint/database/topic_extract.py
FrancescoCasalegno: Thank you for the explanation, I had tried it on my laptop and I did not notice that you indeed get misleading help messages for `choices=ArticleSource`.

I tried setting `ArticleSource.__repr__` but it also does not work! 🤷

src/bluesearch/database/topic_info.py
FrancescoCasalegno: True, even if we got rid of the `metadata` we would still have slightly different data structure in the `json` we want to return.

In fact I was thinking whether there was any real advantage in storing info in a different way in this class and in the `json` we need to use elsewhere. If we had the same structure in those two places then `asdict` would do and would streamline things imo.

But looks good to me as it is as well!

FrancescoCasalegno (approved):
Lgtm!

PR #548, opened 2022-01-17
Fix Postgres issue

### LINKED ISSUE #547: `tests/integration/test_bbs_database.py::test_bbs_database[postgres]` fails randomly
# Description
See below a list of runs where it failed.
* https://github.com/BlueBrain/Search/actions/runs/1687864091/attempts/1
* https://github.com/BlueBrain/Search/actions/runs/1678055422

Unfortunately, it seems to be random and most of the time it passes.

jankrepl:
Fixes #547 

## Description
[ProgrammingError](https://docs.sqlalchemy.org/en/14/core/exceptions.html#sqlalchemy.exc.ProgrammingError) seems to be the issue here so we add it to the list of exceptions that we want to catch.

### DIFF
--- a/tests/integration/test_bbs_database.py
+++ b/tests/integration/test_bbs_database.py
@@ -5,7 +5,7 @@ import docker
 import pg8000
 import pytest
 import sqlalchemy
-from sqlalchemy.exc import InterfaceError, OperationalError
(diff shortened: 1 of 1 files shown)


Stannislav (approved):
LGTM

jankrepl:
> Thanks for fixing this!

Hehe, well, who knows if it will really fix it.

PR #564, opened 2022-02-01
Connecting `topic-filter` and `parse`

### LINKED ISSUE #559: Design an intermediary step between `bbs_database topic-filter` and `bbs_database parse`
## 🚀 Feature

If we don't make any major changes in #550 then the output of `bbs_database topic-filter` is going to be a `.csv` file that looks like this:

```
# output.csv
path,element_in_file,accept,source                                                                                                                                           
path_biorxiv_1,,True,biorxiv                                                                                                            
path_biorxiv_2,,True,biorxiv                                                                                                            
path_medrxiv_1,,True,medrxiv                                                                                                            
path_medrxiv_2,,True,medrxiv
path_pubmed_2,387.0,True,pubmed 
...
```

On the other hand the `bbs_database parse` command expects a single file or folder (+ it supports regex matching).

Clearly, there needs to be some "middle step" that processes the `output.csv` and sends it through to `bbs_database parse`. Some ideas / requirements:
* Only parse those articles where `accept=True`
* Group based on the `source` - ideally we would only call `bbs_database parse` `n_parsers` times.
* Convert the `source` to the `parser`

jankrepl:
Fixes #559 

## Description
* The `bbs_database parse` was adjusted to also support ingesting of piped standard input
  - Before:
  ```bash
  bbs_database parse parser input_path output_dir
  ```
  - After:
   ```bash
  bbs_database parse parser input_path output_dir # still possible
  echo "file_1.py file_2.py" | bbs_database parse parser output_dir # this one works too
  ```
* A shell script (see below an example) can take care of the creation of the string to be piped. Built from `awk` and `clip`.

## How to test?
First of all, create a bunch of empty files

```bash
touch 0.xml a.xml b.xml c.xml d.xml e.xml f.xml g.xml h.xml
```

Then create the following `output.csv`

```bash
path,element_in_file,accept,source                                                                                                                                                                                                                                              
0.xml,,True,arxiv                                                                                                                                                                                                                                                               
a.xml,,False,biorxiv                                                                                                                                                                                                                                                             
b.xml,,True,biorxiv                                                                                                                                                                                                                                                             
c.xml,,True,medrxiv                                                                                                                                                                                                                                                             
d.xml,,False,medrxiv                                                                                                                                                                                                                                                            
e.xml,387.0,True,pubmed                                                                                                                                                                                                                                                         
f.xml,387.0,True,pubmed                                                                                                                                                                                                                                                         
g.xml,387.0,True,pmc                                                                                                                                                                                                                                                            
h.xml,387.0,False,pmc   
```
Then you can run the following script. Note that it does both the grouping based on the `source` and also only takes the articles that were "accepted" in the `topic-filter` stage.


```bash
SOURCES=("arxiv" "biorxiv" "medrxiv" "pubmed" "pmc")                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                
OUTPUT=output.csv                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
for SOURCE in "${SOURCES[@]}"; do                                                                                                                                                                                                                                               
    PARSER='cord19-json'  # to be adjusted                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                
    echo source=$SOURCE                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
    pipe=$(awk -F , '$3 == "True"  && $4 == '\"$SOURCE\"' { print }' $OUTPUT | cut -d ',' -f1)                                                                                                                                                                                  
    echo $pipe | bbs_database parse -n $PARSER final_output_folder                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                
    echo "-------------"                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
done                                                                                                                                                                                                                                               
```
It should give you the following 
```
source=arxiv
0.xml
-------------
source=biorxiv
b.xml
-------------
source=medrxiv
c.xml
-------------
source=pubmed
e.xml
f.xml
-------------
source=pmc
g.xml
-------------
```

## Issues / to discuss
* When we use the `bbs_database parse` in the pipe mode `pdb` does not seem to work
* `pytest` sets stdin to null and `sys.stdin.isatty() == False` which makes the testing more tricky. This PR spawns subprocesses to avoid problems. Unfortunately, these tests are not going to count towards the coverage.
## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -172,7 +172,7 @@ def run(
             raise ValueError("No input files provided")
         else:
             # Piped session
-            input_lines = sys.stdin.read().splitlines()
+            input_lines = sys.stdin.read().split()
             inputs = []
 
             for line in input_lines:
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -153,7 +153,7 @@ def iter_parsers(input_type: str, input_path: Path) -> Iterator[ArticleParser]:
 def run(
     *,
     input_type: str,
-    input_path: Path or None,
+    input_path: Path | None,
     output_dir: Path,
     match_filename: str | None,
     recursive: bool,
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -79,8 +79,8 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         help="""
         Path to a file or directory. If a directory, all articles
         inside the directory will be parsed. If not provided the
-        user is supposed to pipe a list of article paths to the standard
-        input.
+        user is supposed to pipe a space separated list of article paths to
+        the standard input.
         """,
         nargs="?",
     )
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -169,7 +169,8 @@ def run(
     if input_path is None:
         if sys.stdin.isatty():
             # Real terminal session
-            raise ValueError("No input files provided")
+            logger.error("No input files provided")
+            return 1
         else:
             # Piped session
             input_lines = sys.stdin.read().split()
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -20,8 +20,8 @@ from __future__ import annotations
 import argparse
 import json
 import logging
-import warnings
 import sys
+import warnings
 from pathlib import Path
 from typing import Iterator
 
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -21,6 +21,7 @@ import argparse
 import json
 import logging
 import warnings
+import sys
 from pathlib import Path
 from typing import Iterator
 
@@ -77,8 +78,11 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         type=Path,
         help="""
         Path to a file or directory. If a directory, all articles
-        inside the directory will be parsed.
+        inside the directory will be parsed. If not provided the
+        user is supposed to pipe a list of article paths to the standard
+        input.
         """,
+        nargs="?",
     )
     parser.add_argument(
         "output_dir",
@@ -149,7 +153,7 @@ def iter_parsers(input_type: str, input_path: Path) -> Iterator[ArticleParser]:
 def run(
     *,
     input_type: str,
-    input_path: Path,
+    input_path: Path or None,
     output_dir: Path,
     match_filename: str | None,
     recursive: bool,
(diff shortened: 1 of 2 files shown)


Stannislav (approved):
Looks good to me. Can the piped input be of any arbitrary length?

jankrepl:
> Looks good to me. Can the piped input be of any arbitrary length?

Good question! I guess this SO thread answers it https://unix.stackexchange.com/questions/11946/how-big-is-the-pipe-buffer

On my laptop it the limit seems to be 65536 bytes. Assuming the path takes 100 bytes (?=100 chars?) then we could feed 600+ paths at a time. Are my calculations correct? If yes, we can always chunk it up via shell scripting somehow, no?

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -162,14 +166,30 @@ def run(
...
+            return 1
+        else:
+            # Piped session
+            input_lines = sys.stdin.read().split()
+            inputs = []
+
+            for line in input_lines:
+                path = Path(line)
+                if path.exists():
+                    inputs.append(path)
```
EmilieDel, line 182: Do you think we should inform the user when the path does not exist ? Normally it should never happen, right ?

EmilieDel (approved):
LGTM :)

src/bluesearch/entrypoint/database/parse.py
jankrepl: I wondered about the same thing!

FrancescoCasalegno (commented):
Thanks for point out that this intermediate step is needed and for providing this solution @jankrepl !

If we take a step back, what are all the possible solutions and what are the corresponding pros / cons?
1. Modify `parse` to allow a variable number of inputs that will be provided with a pipe `|`.
    - **Pros**
      - Logically simple to understand, command does one and only one thing.
    - **Cons**
       - Piping may break if there are too many files, so manual batching is needed (= extra bash code to chunk the list of input files).
       - External pre-processing with `awk` and `cut` is needed to pre
 2. Modify `parse` to allow two input modes: (a) path to a file or dir to parse (b) path to a topic-info file.
     - **Pros**
        - Easier to use, just one line on the terminal w/o need for pre-processing with `awk`.
     - **Cons**
        - Interface may result less clear (two different ways to call the command).
        - We have to implement the parsing logic on a for loop over the files mentioned in the input `.csv` file. Basically we are moving user code to be written in a `bash` script (option 1) into the command (this option 2).
        

What is your feelings on which one would be best? @jankrepl  @Stannislav @EmilieDel

jankrepl:
> Thanks for point out that this intermediate step is needed and for providing this solution @jankrepl !
> 
> If we take a step back, what are all the possible solutions and what are the corresponding pros / cons?
> 
> 1. Modify `parse` to allow a variable number of inputs that will be provided with a pipe `|`.
>    
>    * **Pros**
>      
>      * Logically simple to understand, command does one and only one thing.
>    * **Cons**
>      
>      * Piping may break if there are too many files, so manual batching is needed (= extra bash code to chunk the list of input files).
>      * External pre-processing with `awk` and `cut` is needed to pre
> 2. Modify `parse` to allow two input modes: (a) path to a file or dir to parse (b) path to a topic-info file.
>    
>    * **Pros**
>      
>      * Easier to use, just one line on the terminal w/o need for pre-processing with `awk`.
>    * **Cons**
>      
>      * Interface may result less clear (two different ways to call the command).
>      * We have to implement the parsing logic on a for loop over the files mentioned in the input `.csv` file. Basically we are moving user code to be written in a `bash` script (option 1) into the command (this option 2).
> 
> What is your feelings on which one would be best? @jankrepl @Stannislav @EmilieDel

I don't love either of the solutions. However, I still prefer 1. since it does not expand the scope of `parse`. Regarding the chunking, I guess we will have to wait and see how big of a problem it will be on real data.

PR #570, opened 2022-02-04
Address `bbs_database topic-extract` inconsistency unzipping for different sources

### LINKED ISSUE #551: `bbs_database topic-extract` inconsistent behavior for different sources - unzipping
## Description
Current logic

* `source=pubmed`: `bbs_database download` -> `.xml.gz` file -> manual unzipping (`gzip -d  *.xml.gz`) ->  `.xml` file--> `topic-extract`
* `source=biorxiv/medrxiv`: `bbs_database download` -> `.meca` file -> `topic-extract` (unzipping implemented inside) 

## Possible solutions
1. No action. We just need to make sure that once we design the pipeline script we take it into account.
2. Remove unzipping logic from `source=biorxiv/medrxiv`. IMO this one would be a bad idea since after unzipping there are multiple files and one needs to programatically find a unique `.xml` among them.
3. Add unzipping logic to `source=pubmed`. Might make the code + tests a little bit less readable.

EmilieDel:
Fixes #551.

## Description

## How to test?

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/src/bluesearch/entrypoint/database/topic_extract.py
+++ b/src/bluesearch/entrypoint/database/topic_extract.py
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import argparse
+import gzip
 import logging
 from pathlib import Path
 from typing import Any
@@ -160,7 +161,10 @@ def run(
     elif article_source is ArticleSource.PUBMED:
         for path in inputs:
             logger.info(f"Processing {path}")
-            articles = ElementTree.parse(input_path)
+            with gzip.open(path) as mygzip:
+                articles_str = mygzip.read()
+            articles = ElementTree.fromstring(articles_str)
+
             for i, article in enumerate(articles.iter("PubmedArticle")):
                 topic_info = TopicInfo(
                     source=article_source,
--- a/tests/unit/entrypoint/database/test_topic_extract.py
+++ b/tests/unit/entrypoint/database/test_topic_extract.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 import argparse
+import gzip
 import inspect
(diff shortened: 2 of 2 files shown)


jankrepl (approved):
Great! Thank you!

src/bluesearch/entrypoint/database/topic_extract.py
```diff
@@ -160,7 +161,10 @@ def run(
     elif article_source is ArticleSource.PUBMED:
         for path in inputs:
             logger.info(f"Processing {path}")
-            articles = ElementTree.parse(input_path)
+            with gzip.open(path) as mygzip:
+                articles_str = mygzip.read()
+            articles = ElementTree.fromstring(articles_str)
+
```
FrancescoCasalegno, line 167: So after this we are assuming in `topic-extract` that the input is a `.xml.gz` file when the source is `PubMed`.
But in `parse` don't we assume we have the `xml` file (i.e. that the `.gz` was unzziped)?
https://github.com/BlueBrain/Search/blob/6e362760ef0d2bbb6cd583dfb363313d222737e2/src/bluesearch/entrypoint/database/parse.py#L130-L136

Maybe this is more a question for @jankrepl who opened #551.

EmilieDel: I think that as soon as we can do without unzipping and use directly the raw files, it is better (less intermediate/manual) steps. If there are other inconsistencies, we should maybe address them. Do you want me to make the changes in this in the current PR?

FrancescoCasalegno: I am fine leaving this part outside of the current PR, but maybe just open an issue with a reference to this PR so we can address it next :)

PR #571, opened 2022-02-08
Luigi pipeline sketch

### LINKED ISSUE #558: `bbs_database parse pubmed-xml-set` - impossible to specify element in file
## 🚀 Feature

Currently, if one uses the `pubmed-xml-set` in `bbs_database parse` then **every** single article in the article set is going to be parsed. However, it might be useful (e.g. only keeping relevant topics) to select a subset of articles (elements) in the article set to parse.

Note that this logic would be only relevant for `pubmed-xml-set` parser and we need to make sure that we don't clutter the `bbs_database parse` too much

### LINKED ISSUE #562: First sketch of the overall pipeline (+ research relevant tools/approaches)
## 🚀 Feature
We are getting closer and closer to implementing all the relevant steps of our ETL pipeline. IMO it might be beneficial to create a first sketch of the overall pipeline and also look into tools that might be relevant.

Some ideas/to discuss
a) What logic/tool do we use to define the pipeline (raw shell script, DVC-like tools, ...)
b) How do we trigger it? (cronjob, manually, ...)
c) How do we monitor (live ideally) the progress of a running pipeline (something like github actions or jenkins do)
d) How do we test it (e.g. extending current integration tests - related #532)

jankrepl:
Fixes #562 and fixes #558

### Diagrams
Folder naming logic
Screenshot 2022-02-21 at 14 04 51

Pipeline steps
Screenshot 2022-02-18 at 11 56 47

- diagram source - https://pastebin.com/raw/5KhePmK0 (just copy and paste in tldraw)
### Important design choices
* The CLI has no positional arguments and instead it is using the concept of required options
* The outputs (as shown in the diagram) will be put in the following folder `$output_dir/$source/$date/`
   * `output_dir` is specified via the CLI (`--output-dir`)
   * `source` is also specified via the CLI
   * `date` is a concatenation of the below two dates (using the `_` separator for joining) (e.g. `2022-02_2022-02-15`):
     * `from-month` (format `%Y-%m`) (specified via the CLI)
     * today (format `%Y-%m-%d`)
* The `PerformFilteringTask` creates symlinks to those articles that made it through the filtering and puts these symlinks to `filtered/`. This makes it very easy for `ParseTask` to ingest them.
* The mesh database needs to be provided via the CLI via `--mesh-topic-db=mesh_topic_db.json` -> The `bbs_database parse-mesh-rdf` is not included in the pipeline.
* The code disables the central scheduler (via `local_scheduler=True`). It is mostly for debugging purposes. However, we might want to use the central scheduler in "production". See more info https://luigi.readthedocs.io/en/stable/central_scheduler.html 
* Luigi can be very verbose. Its logging level was hardcoded to `WARNING`. Feel free to modify it to have more information on what `luigi` is doing.
## How to test?
First of all, initialize an empty database.

```bash
bbs_database init my_db.sqlite
```
Make sure you define the topic rules in `rules_config.jsonl`. If you want all articles to be accepted (debugging) just do
```json
{"label": "accept"}
```

You can create a helper script `launch_luigi.sh`.
```bash
 bbs_database run \
     --source=pubmed \
     --from-month=2022-02 \
     --filter-config=rules_config.jsonl \
     --output-dir=pipeline_results/ \
     --mesh-topic-db=mesh_topic_db.json \
     --db-url=my_db.sqlite \
     --grobid-host=dgx1.bbp.epfl.ch \
     --grobid-port=8070 \
     $@
```
And you can then run `luigi` in 2 different ways
**Dry run**
```bash
./launch_luigi.sh -n
```
* Shows what steps are still to do (`PENDING`) and which are done (`COMPLETED`).

**Real run**
```bash
./launch_luigi.sh
```

To speed up the pipelines (debugging purposes) consider creating the following script`custom_timeout`.
```bash
timeout 15 $@ || ( [[ $? -eq 124 ]] && exit 0 )
```
You can then simply use it inside of `ExternalProgramTask` as the first argument. Very useful for `DownloadTask`. Make sure that the Python process can find it in the `PATH` environment variable.

## TODO + To discuss
- [x] Write end to end tests (e.g. make the raw download as real as possible). Think of a way how to run "test" runs without having to download everything (maybe `timeout` but need to handle the nonzero exit code somehow)
   -  The best solution was to timeout the `DownloadTask` so that there are not that many articles
- [x] Handle logging and verbosity (`luigi` vs our entrypoints)
   - For now we just automatically set the logging level of all entrypoints to `logging.INFO`. However, can be changed using the hardcoder "global" variable `VERBOSITY`
- [x] Make sure no unzipping needed for `pubmed` (there is a PR for this)
   - [x] No unzipping needed for `topic-extract`
   - [x] No unzipping needed for `parse`
- [x] Verify that `biorxiv` and `medrxiv` work end-to-end
- [x] Best way to provide optional parameters to our entrypoints (e.g. `luigi.cfg`)
   - Let's do this in a different PR
   - e.g. the number of processes for `convert-pdf`
- [x] Test on `--db-type=postgres`
- [x] `bbs_database run --dry-run` remove the parameters to make it more readable
   - The parameters are important. 
- [x] Should we replace `ExternalProgramTask` with simple `from bluesearch.entrypoint.database.? import run; run()`? Maybe we don't need to replace it at all.
- [ ] Refactor
- [x] Document
- [x] Try to make `luigi` history work
   - Let's do this in a different PR if useful 
- [x] What happens if the pipeline does not run within 1 day?
  - Solved via a global variable that stores a unique folder for the entire process 
- [x] Add `luigi` to requirements
- [ ] Test with more realistic `rules.json` 
- [ ] Make it possible to stop early

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [x] All CI tests pass.


EmilieDel (approved):
Amazing PR! 🎉

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,572 @@
...
+
+        else:
+            raise ValueError(f"Unsupported source {self.source}")
+
+
+@inherits(DownloadTask, UnzipTask)
+class TopicExtractTask(ExternalProgramTask):
+    """Topic extraction.
+
+    The input of this dask is either `raw/` or `raw_unzipped/` depending
```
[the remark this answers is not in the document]
jankrepl: Done in 7c50ca4

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,572 @@
...
+        filtering = pd.read_csv(self.input().path)
+
+        output_dir.mkdir(exist_ok=True)
+
+        if self.source == "pubmed":
+            # Find all input files (.xml.gz)
+            all_input_files = [Path(p) for p in filtering["path"].unique()]
+
+            # Iteratively Load each  of the files in memory
+            for input_file in all_input_files:
```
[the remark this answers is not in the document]
jankrepl: IMO it should be fine since these paths are coming from the `filtering.csv` table. And it was generated by `topic-filter` so I would assume that are valid paths.

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,572 @@
...
+
+        return luigi.LocalTarget(str(output_file))
+
+
+@inherits(ConvertPDFTask, PerformFilteringTask)
+class ParseTask(ExternalProgramTask):
+    """Parse articles.
+
+    The input is all the articles inside of `filtered/` (or in case of
+    `source="arxiv"` `converted_pdfs/`.
```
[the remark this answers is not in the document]
jankrepl: 927fc89

src/bluesearch/entrypoint/database/run.py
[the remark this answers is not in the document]
EmilieDel: True! :)

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,636 @@
...
+        all_dependencies_map = {t.__class__.__name__: t for t in all_dependencies}
+
+        if final_task in all_dependencies_map:
+            selected_task_inst = all_dependencies_map[final_task]
+        else:
+            raise ValueError(f"Unrecognized final task {final_task}")
+
+    luigi_kwargs = {
+        "tasks": [selected_task_inst],
+        "log_level": "WARNING",
```
FrancescoCasalegno, line 627: Would it make sense instead to look at the the verbosity level that is passed on the command line to the `bbs_database` command and apply it to `luigi` as well, instead of setting a default?

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,636 @@
...
+    else:
+        all_dependencies = get_all_dependencies(add_task_inst)
+        all_dependencies_map = {t.__class__.__name__: t for t in all_dependencies}
+
+        if final_task in all_dependencies_map:
+            selected_task_inst = all_dependencies_map[final_task]
+        else:
+            raise ValueError(f"Unrecognized final task {final_task}")
+
+    luigi_kwargs = {
```
FrancescoCasalegno, line 625: An alternative could be to have those `luigi_conf` as default, but allow user to overwrite them on the CLI by passing a list of key-value pairs that will be passed to luigi, like this:
```bash
bbs_database run ... --luigi-config=log_level:WARNING,local_scheduler:True,...
```
Especially because otherwise changing things like the scheduler (local/central) would require modifying every time the source code, which IMO is not ideal...

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,636 @@
...
+            *VERBOSITY,
+            self.db_url,
+            input_dir,
+            f"--db-type={self.db_type}",
+        ]
+
+        return command
+
+
+def get_all_dependencies(task: luigi.Task) -> set[luigi.Task]:
```
FrancescoCasalegno, line 546: If I understand correctly, you are using this function only to retrieve the class name of the `final_task`?
But then wouldn't it be easier to just do something like `getattr` to get the class object from its `str` name?

FrancescoCasalegno (commented):
Hi @jankrepl , thank you for this immense work!
I added a few comments, but I think the overall structure looks great!

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,636 @@
...
+    return parser
+
+
+BBS_BINARY = ["bbs_database"]
+VERBOSITY = ["-v"]  # for the entrypoint subprocesses
+CAPTURE_OUTPUT = False
+IDENTIFIER = None  # make sure the same for all tasks
+
+
+class DownloadTask(ExternalProgramTask):
```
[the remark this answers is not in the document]
jankrepl: We discussed this extensively with @EmilieDel and we actually went "full circle" when working on this  PR: `ExternalTask` -> `Task` -> `ExternalTask`. 

Let me point out some of the motivation behind using `ExternalTask`
* The `run` functions assume the input parameters have been parsed by the corresponding parser. So in order to use these `run` functions directly one needs to make sure the types are correct (=extra work). By using `ExternalTask` the parsing is done automatically.
* The startup time of `bbs_database command` is fast and there does not seem to be any major overhead of creating a subprocess
* In theory, the pipeline steps could be run in a different virtual env/docker container/image than this `luigi` pipeline. Of course this would require some extra work but just the logical separation that the pipeline runner process is different from the worker process felt like the right choice.
* This PR is a just a sketch.

src/bluesearch/entrypoint/database/run.py
jankrepl: Definitely something to be discussed. IMO there are 3 "namespaces" that we can control the verbosity of.


1) `luigi` itself
2) `bbs_database run` - basically all the `logger.info/logger.debug` inside of `entrypoint/database/run.py`
3) Subprocesses - basically all the `ExternalProgramTask` that are calling `bbs_database command`

The current status is that
1) Controlled via `bbs_database run -v`
2) Hardcoded to `WARNING`
3) Hardcoded to `-v` for all subprocesses

You are suggesting to merge 1) + 2) and IMO it is not good idea since `luigi` can be extremely verbose when the level is `INFO` or `DEBUG` and it might actually make it harder to "debug" problems in our code just by looking at the standard output.

src/bluesearch/entrypoint/database/run.py
jankrepl: I like the idea;)

Again, this PR was just a sketch and all the luigi configuration was hardcoded to whatever was the most convenient when developing. IMO we could even do this in a separate PR since it might be worth looking into what the best practices are when it comes to providing "dictionary-like" data via CLI.

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,636 @@
...
+        output_dir = Path(self.output().path)  # raw_unzipped
+
+        output_dir.mkdir(exist_ok=True, parents=True)
+        if self.source == "pmc":
+            # .tar.gz
+            # We want collapse the folder hierarchy
+            all_tar_files = input_dir.rglob("*.tar.gz")
+            for archive in all_tar_files:
+                output_path = output_dir / archive.stem
+                my_tar = tarfile.open(archive)
```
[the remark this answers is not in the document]
jankrepl: Done in 7c81595

src/bluesearch/entrypoint/database/run.py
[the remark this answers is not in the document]
FrancescoCasalegno: ## Discussed on 2022-03-08

Looks fine with `ExternalTask` for the time being.

---- 
> The `run` functions assume the input parameters have been parsed by the corresponding parser. So in order to use these `run` functions directly one needs to make sure the types are correct (=extra work). By using `ExternalTask` the parsing is done automatically.

As a note, this could also be achieved by calling, e.g., 
```python
bluesearch.entrypoint.database.parent.main("download", "arg1", ...)
```
instead of 
```python
bluesearch.entrypoint.database.download.run(arg1, ...)
```

src/bluesearch/entrypoint/database/run.py
FrancescoCasalegno: I tested now, and the configuration can indeed be moved away from the source code.
Just modify the `luigi.cfg` to look as follows:
```python
[core]
    autoload_range = True
    log_level = DEBUG
    local_scheduler = True
```

jankrepl: Done in `8830e69`. I set the logging level to `INFO` since it seems to be a reasonable default.

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,633 @@
...
+
+@requires(DownloadTask)
+class UnzipTask(ExternalProgramTask):
+    """Unzip raw files (if necessary).
+
+    Only applicable in case of `pmc`. The unzipped files
+    are stored inside of `raw_unzipped`.
+    """
+
+    source = luigi.Parameter()
```
FrancescoCasalegno, line 214: Don't we already have
```python
@requires(DownloadTask)
```
a few lines before?

jankrepl: I am not sure what you are referring to, however, `source` is going to be `arxiv`, `medrix`, etc

jankrepl: Ahh, I think I get your point now. Yes, the `requires` is going copy paste all the parameters of the `DownloadTask`. So it is not necessary to specify it again, however, we just though it would be nice to define it explicitly again those parameters that are necessary for that given task. But it is definitely something we can discuss.

src/bluesearch/entrypoint/database/run.py
FrancescoCasalegno: > No, this function gets all task instances.

Sure, this was clear. I was referring to what the actual usage of it in the code – namely the fact that "we only select the one that the user specified as `--final-task`".


----

> The approach your suggested would require passing all the parameters necessary to instantiate a given class.

Ah now I see the issue, thanks! Anyway, notice that `luigi` allows to provide the `Parameter` values of any `Task`  in the `luigi.cfg`. This means that we can simply execute 
```python
# Note: I am calling __init__() w/o defining values for the parameters here!!
luigi.build(tasks=[AddTask()])  
```
as long as we have a `luigi.cfg` that looks like this one:
```
[DownloadTask]
source = arxiv
from_month = 2021-12
...

[GenerateWordsTask]
...

[AddTask]
...
```

Now, there is only one small issue: since you have used `@inherits` and `@requires` everywhere, `luigi` expects **every** `Task` we have defined in our pipeline to have those parameters set somewhere. If we use the aforementioned approach with the parameters set in `luigi.cfg`, we have two possibilities.
1. If `AddTask` is the `final_task`, then we can just define all parameters under `[AddTask]`.
2. If the `final_task` is something else that we don't know in advance (our case), then we need every `[XXXTask]` to have defined all the inherited parameters. For instance, every `[XXXTask]` will need to define `source = ...`!

I think that our use case is option 2., but the fact of having to define `source = ...` many times in the `luigi.cfg` is not really great. Anyway, I think that this is not really `luigi`'s fault, it's more the fact we have those `@inherits` and `@requires` everywhere, even though I am not sure that we should really do that — why setting `@requires(ParseTask)` for `AddTask`, as the latter does not need any of the parameters of the former?

FrancescoCasalegno: Sorry for the lengthy comment, here's a summary of the alternative I see for the current approach.

➡️ (Unless we really need to re-use parameters in downstream tasks) get rid of the `@inherits` and `@requires` + set `Parameter` values of each `Task` in the corresponding section in the `luigi.cfg`.

And here's the main benefits.

😊 Use a configuration file to configure the parameters of each `luigi` `Task` seems to be cleaner. Also, no need to crawl the `Task` dependency graph to retrieve the various instances. Finally, by getting rid of unneeded `@inherits`/`@requires`, each `Task` only has attached to it the `Parameter`s it really needs.

FrancescoCasalegno: We tested it, and we can indeed use empty `__init__()` for any `Task`, as long as the `luigi.cfg` defines values for each required `Parameter`.

Still to do:
- [ ] what happens if the `luigi.cfg` is modified _after_ the `luigi.build()` is called? is it re-loaded? or ignored? (especially after it starts a new task)
- [ ] is there any way to specify only once the value of `source` which is shared by many `Task`s? e.g. maybe in the `luigi.cfg`? or as a dummy `Task`?

EmilieDel: From my experiments today, it seems:
- The `luigi.cfg` configuration file is not reloaded during a run. It means that even if the file is modified while luigi pipeline is running, the initial parameters are kept.
- The way to handle shared parameters is through a `luigi.Config` class containing the global parameters. See (here)[https://stackoverflow.com/questions/48509083/how-to-make-a-parameter-available-to-all-luigi-tasks]

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,633 @@
...
+@inherits(DownloadTask, UnzipTask)
+class TopicExtractTask(ExternalProgramTask):
+    """Topic extraction.
+
+    The input of this task is either `raw/` or `raw_unzipped/` depending
+    on the source. The output is going to be a single file
+    `topic_infos.jsonl`.
+    """
+
+    source = luigi.Parameter()
```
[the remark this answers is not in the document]
FrancescoCasalegno: This comment of mine is not relevant anymore, as we are getting rid of the `@inherits`.

src/bluesearch/entrypoint/database/run.py
```diff
@@ -0,0 +1,548 @@
...
+
+        config = luigi.configuration.get_config()
+        config.add_config_path(config_path)
+        config.reload()
+
+    if luigi_config:
+        config = luigi.configuration.get_config()
+        for param in luigi_config.split(","):
+            change = re.split(r"[.:]", param, maxsplit=3)
+            config.set(*change)
```
FrancescoCasalegno, line 536: Do we need both `config-path` and `luigi-config` CLI arguments? I am not sure I understand the difference.

Also, isn't it easier/better to remove them and just rely on the env variable `LUIGI_CONFIG_PATH` (see [`luigi` Docs](https://luigi.readthedocs.io/en/stable/configuration.html#configuration))? This is for instance what BBP's `mcar` does, see e.g. how we are invoking it [here ](https://bbpgitlab.epfl.ch/ml/deepmorphs/-/blob/main/dvc/dvc.lock#L9149) in another project.
Then we would be able to run the entrypoint like this:
```bash
LUIGI_CONFIG_PATH=some/path/luigi.cfg bbs_database run [args...]
```

luigi.cfg
```diff
@@ -0,0 +1,27 @@
...
+[TopicFilterTask]
+    filter_config=luigi-pipeline/filter-config.jsonl
+
+[ConvertPDFTask]
+    grobid_host=0.0.0.0
+    grobid_port=8070
+
+[AddTask]
+    db_url=luigi-pipeline/my-db.db
+    db_type=sqlite
```
FrancescoCasalegno, line 27:
```
;Blue Brain Search is a text mining toolbox focused on scientific use cases.
;
;Copyright (C) 2020  Blue Brain Project, EPFL.
;
;This program is free software: you can redistribute it and/or modify
;it under the terms of the GNU Lesser General Public License as published by
;the Free Software Foundation, either version 3 of the License, or
;(at your option) any later version.
;
;This program is distributed in the hope that it will be useful,
;but WITHOUT ANY WARRANTY; without even the implied warranty of
;MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
;GNU Lesser General Public License for more details.
;
;You should have received a copy of the GNU Lesser General Public License
;along with this program. If not, see <https://www.gnu.org/licenses/>.

[core]
autoload_range = true
log_level = INFO
local_scheduler = True

[GlobalParams]
source = pubmed

[DownloadTask]
from_month = 2021-12
output_dir = luigi-pipeline
identifier =
; emtpy string is considered default value

[TopicExtractTask]
mesh_topic_db = luigi-pipeline/mesh_topic_db.json

[TopicFilterTask]
filter_config = luigi-pipeline/filter-config.jsonl

[ConvertPDFTask]
grobid_host = 0.0.0.0
grobid_port = 8070

[AddTask]
db_url = luigi-pipeline/my-db.db
db_type = sqlite

```
INI/CFG linting + BBP header.

FrancescoCasalegno (commented):
LGTM, amazing work with the refactoring!

I left two minor suggestions, then feel free to merge it for real 🔥

luigi.cfg
EmilieDel: Thank you! Done in [bd77681](https://github.com/BlueBrain/Search/pull/571/commits/bd77681a87f42973760646f19330e053e4b1b712).

src/bluesearch/entrypoint/database/run.py
EmilieDel: - `config-path` is there to change the path to the configuration file. 
- `luigi-config` is to allow the user to overwrite a parameter (if needed). e.g. `--luigi-config GlobalParams.source:arxiv`.

If you prefer, I can remove those two arguments. Maybe just a side note related to this: I have the feeling it is much easier to tests the complete pipeline with those two arguments.

FrancescoCasalegno: Suggestions:
- [ ] Check what happens if a user provides `config-path` but then `LUIGI_CONFIG_PATH` is also set as an environmental variable. What has precedence?
- [ ] Rename `config-path` as `luigi-config-path` + make parser helper more explicit (`help="Configuration Path."` --> `help="Path to Luigi configuration file."`)
- [ ] Rename `luigi-config` as `luigi-config-args`  + make parser helper more explicit (`help="Configuration parameters."` --> `help="Comma separated key-value arguments for Luigi configuration, e.g. "--luigi-config GlobalParams.source:arxiv,DownloadTask.from-month:2021-04" . Overwrites the content of Luigi configuration file (see --luigi-config-path)."`)

EmilieDel: Done in [0d343d3](https://github.com/BlueBrain/Search/pull/571/commits/0d343d3af274de31c082bb00259dd1f9a70864f1).
For info, the `luigi-config-path`has precedence over the environment variable.

PR #589, opened 2022-02-25
Log info on processed batch for `topic-extract` on `arxiv` 

FrancescoCasalegno:
## Description

Currently, we are logging info while we extract topic info for all sources but `arxvi`. 

The reason why `arxiv` is special is that we don't send requests to the server using the `arxiv ids` one by one, but in batches.

So this PR introduces batch-based logging info for `arxiv` topic extraction.

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.


jankrepl (approved):
LGTM

EmilieDel (approved):
LGTM! Thank you!

PR #628, opened 2022-09-15
First draft of adding topic to json

### LINKED ISSUE #619: Include topic info in JSON output parse of `bbs_database parse`
## Context
- We have implemented the commands `bbs_database topic-extract` to extract information about the topics of an article.
- However, this topic information is only used by `bbs_database topic-extract` to determine, based on user-defined inclusion/exclusion criteria, whether or not an article file should be parsed into a JSON (with `bbs_database parse`) and then uploaded into the database.
- We would like this information to be preserved also into the `articles` table, so that later on we can e.g. filter articles by topic.

## Actions
- [ ] Include information on the topic in the JSON output parse of `bbs_database parse`.

EmilieDel:
Fixes #619.

## Description

## How to test?

## Checklist

- [ ] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [ ] Type annotations added.
- [ ] All CI tests pass.

### DIFF
--- a/src/bluesearch/entrypoint/database/parse.py
+++ b/src/bluesearch/entrypoint/database/parse.py
@@ -113,6 +113,14 @@ def init_parser(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         """,
     )
     parser.add_argument(
+        "-i",
+        "--include-topic",
+        action="store_true",
+        help="""
+        If True, include topic inside the parsed json.
+        """,
+    )
+    parser.add_argument(
         "-n",
         "--dry-run",
         action="store_true",
@@ -164,6 +172,7 @@ def run(
     output_dir: Path,
     match_filename: str | None,
     recursive: bool,
+    include_topic: bool,
     dry_run: bool,
 ) -> int:
     """Parse one or several articles.
@@ -171,6 +180,8 @@ def run(
     Parameter description and potential defaults are documented inside of the
     `get_parser` function.
     """
+    import json
+
     from bluesearch.utils import find_files
 
     if input_path is None:
@@ -211,7 +222,7 @@ def run(
         try:
             parsers = iter_parsers(input_type, input_path)
 
-            for parser in parsers:
+            for i, parser in enumerate(parsers):
                 article = Article.parse(parser)
                 output_file = output_dir / f"{article.uid}.json"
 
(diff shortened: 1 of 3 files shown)


src/bluesearch/entrypoint/database/parse.py
```diff
@@ -211,14 +222,30 @@ def run(
...
                 if output_file.exists():
                     raise FileExistsError(f"Output '{output_file}' already exists!")
                 else:
                     serialized = article.to_json()
+
+                    if include_topic:
+                        topic_path = (
+                            input_path.parent.parent
+                            / "topic"
+                            / f"{input_path.stem}.json"
```
jankrepl, line 238: I am not a huge fan of this hardcoding. 

IMO we should not assume this is run within a pipeline.

Why don't we just replace `--include-topic` with `--topic-path` (or similar) and if the user does not provided then it means that they are not interested

jankrepl (commented):
I added one comment.


Also, IMO it would be nice to add some unit tests to make sure things work as expected.

src/bluesearch/entrypoint/database/parse.py
EmilieDel: Yes that is definitely my plan! I did not want to hardcode the variable. I just wanted to be sure we can move on in this direction. What do you think about the way to solve this issue ?

src/bluesearch/entrypoint/database/parse.py
```diff
@@ -211,14 +222,30 @@ def run(
...
+                        )
+                        topic_json = JSONL.load_jsonl(topic_path)
+
+                        serialized_json = json.loads(serialized)
+                        if input_type == "pubmed-xml-set":
+                            serialized_json["topics"] = topic_json[i]["topics"]
```
jankrepl, line 244: So the idea here is to dynamically add this `topics` key only inside of the JSON?

Wouldn't it make more sense to define a new attribute `topcis` for `Article` that is an empty list by default? And then modify this attribute once we have the topics and then serialize the whole article?


This way the schema of the `Article` is going to be clear.

src/bluesearch/entrypoint/database/parse.py
jankrepl: So maybe I can review again once you have a final version?

EmilieDel: Definitely! I let you know when the PR is done :) Thank you for your help!

src/bluesearch/entrypoint/database/parse.py
EmilieDel: I like the idea, but I have the feeling that parsing the article and finding the topic are two completely separate steps. I am a bit afraid of making those steps more dependent on each other. Currently, I try to keep everything independent (and the changes in the CI command reasonable). Do you think it is possible to add the new attribute `topics` in `Article` and keeping things independent?

jankrepl: I see your point and I guess that is why I suggested to set it to `list[str]` by default. If the user wants, they can provide itlater but at least we declare that the `topics` attribute could be populated.

IMO one should just look at the `Article` dataclass and be able to see what the final schema of the saved `.json` is going to be. Once we start writing code that adds new entries to the `.json` then I am worried about things becoming intractable.

Anyway, what do you think?

PR #638, opened 2022-09-27
Remove python 3.7 from CI

### LINKED ISSUE #636: Remove Python 3.7 from the CI
Mainly because it seems like we can have up to 5 parallel runners at a time. Unfortunately, we have 2 OS and 3 Python versions (=6 runners) currently. This results in ~10 minutes of extra waiting time for the CI.

EmilieDel:
Fixes #636.

## Note

From what I read on internet, I understand that a public repository has 20 concurrent jobs (see [here](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits)) but only 5 `macos` jobs (so it might be still a good idea to remove some python versions for `macos`).

## Checklist

- [x] Unit tests added.
  (if needed)
  (if needed)
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.9, 3.10]
+        python-version: ["3.8", "3.9", "3.10"]
         include:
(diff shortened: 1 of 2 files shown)


drsantos89 (approved):
LGTM ;)

PR #643, opened 2022-10-26
add ner k8s

drsantos89:
## Description

Adds a function to perform and store the output of NER.
NER is run remotely using the deployment on Kubernetes. It supports both ML and RULE-based approaches.
The NER output and model version are stored in ES.

## Notes

- One function, `handle_conflits`, is currently found in two repositories (this current PR and one repo on GitLab). It might be interesting to keep it only in `BlueSearch` and import it into the other repository.
- The remote models are relatively slow when running 1 sample at a time (~1-2 paragraphs/s for ML and ~15 for RULE). Testing the remote model speed with `locust` revealed a maximum possible performant of ~15 and ~50 for ML and RULE models, respectively. A multiprocessing option was hence added.
- pool.apply_async compies the arguments to a new memory location. The `client` object is not serializable and needs to be called inside the function if required.
- by default, the function only updates the paragraphs which are empty or with an outdated model version. There is an option to force the update of every paragraph.
- The JSON output of both models is saved in an ES field of type `flattened`. ("This data type can be useful for indexing objects with a large or unknown number of unique keys. Only one field mapping is created for the whole JSON object, which can help prevent a [mappings explosion (https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#mapping-limit-settings) from having too many distinct field mappings.")

## How to test?

tests/unit/k8s/test_ner.py

## Checklist

- [x] Unit tests added.
  (if needed)
- [x] Type annotations added.
- [x] All CI tests pass.


src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+    ----------
+    text
+        Text to perform NER on.
+    article_id
+        Id of the article.
+    paragraph_id
+        Id of the paragraph.
+    ml_model
+        Name of the ML model to use.
+    """
```
EmilieDel: We need to update the docstring :)

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+    Parameters
+    ----------
+    text
+        Text to perform NER on.
+    article_id
+        Id of the article.
+    paragraph_id
+        Id of the paragraph.
+    ruler_model
+        Name of the entity ruler model to use.
```
EmilieDel: Same here :)

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+        Method to use to perform NER.
+    """
+
+    # get paragraphs without NER unless force is True
+    if force:
+        query: dict[str, Any] = {"match_all": {}}
+    else:
+        query = {"bool": {"must_not": {"exists": {"field": "ner"}}}}
+    paragraph_count = client.count(index=index, query=query)["count"]
+    logger.info("There are {paragraph_count} paragraphs without embeddings")
```
EmilieDel: I think we should change this for "NER"

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+        Name of the ES index.
+    ner_method
+        Method to use to perform NER.
+    """
+
+    # get paragraphs without NER unless force is True
+    if force:
+        query: dict[str, Any] = {"match_all": {}}
+    else:
+        query = {"bool": {"must_not": {"exists": {"field": "ner"}}}}
```
EmilieDel: Now that we changed the logic with 2 fields: `ner_ml` and `ner_ruler`. We should also change the query.

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+        unit=" Paragraphs",
+        desc="Updating NER",
+    )
+    for hit in scan(client, query={"query": query}, index=index):
+        if ner_method == "both":
+            results_ml = run_ml_ner(
+                hit["_source"]["text"],
+                os.environ["BENTOML_NER_ML_URL"]
+            )
+            results_ruller = run_ruler_ner(
```
EmilieDel: There is a small typo

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            raise ValueError(f"Unknown NER method: {ner_method}")
+
+        progress.update(1)
+        logger.info(f"Updated NER for paragraph {hit['_id']}, progress: {progress.n}")
+
+    progress.close()
+
+
+def run_ml_ner(text: str, url: str) -> list[dict]:
```
EmilieDel: I am wondering now that the URL is a parameter of the function. Are there any difference between `run_ml_ner` and `run_ml_ruler` ? Maybe we can merge those two functions. What do you think ?

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
+# the Free Software Foundation, either version 3 of the License, or
```
EmilieDel, line 7: Small detail, but is this the beginning of the license header ?

src/bluesearch/k8s/ner.py
```diff
@@ -0,0 +1,204 @@
...
+    """Runs the NER pipeline on the paragraphs in the database.
+
+    Parameters
+    ----------
+    client
+        Elasticsearch client.
+    index
+        Name of the ES index.
+    ner_method
+        Method to use to perform NER.
```
EmilieDel, line 62: We are missing the `force` parameter in the docstring

